PR #15, opened 2023-09-28
Move interpolation resolving to very end

cleverguy25:
This delays interpolation resolution to the very end to allow overriding values to behave as **I think** customers would expect.

This should resolve two issues, the first one being what I think customers would perceive as a bug with an env hierarch like this:
Env A
```yaml
values:
   first: hello
   second: world
   test: ${first}, ${second}
```
Env B
```yaml
imports:
- a
values:
   second: everybody
```

I believe most people will expect `test` to have the value "hello, everybody" instead of "hello, world"

We may want to have some sort of marker/option to override this behavior, but that is easily setup with separate variables as well.

This would also solve https://github.com/pulumi/pulumi-service/issues/15620  allowing for this config to work:
Env A
```yaml
values:
   test: https://${serviceName}.${aws.region}.${environmentName}.contoso.com
```
Env B
```yaml
imports:
- a
values:
   aws:
      region: us-west
   environmentName: prod
   serviceName: myservice
```

### DIFF
--- a/eval/eval.go
+++ b/eval/eval.go
@@ -98,6 +98,7 @@ func evalEnvironment(
 
 	ec := newEvalContext(ctx, validating, name, env, providers, envs, map[string]*value{})
 	v, diags := ec.evaluate()
+	ec.resolveInterpolations(v)
 
 	s := schema.Never().Schema()
 	if v != nil {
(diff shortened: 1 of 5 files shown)


pgavlin:
Tracked by https://github.com/pulumi/esc/issues/127.

PR #117, opened 2023-10-13
eval: add fn::from{Base64,JSON}

### LINKED ISSUE #103: consider adding a `fromJSON` function
awssecretsmanager recommends storing a json blob as the secret but only returns a string. we need a way to get the object from the json.

### LINKED ISSUE #104: consider adding a `fromBase64` function

pgavlin:
These changes add two new builtins, `fn::fromBase64` and `fn::fromJSON`. The former decodes a base64-encoded string into a binary string and the latter decodes a JSON string into a value.

Fixes #103.
Fixes #104.


pgavlin:
helps to add all the files 🤦

komalali (approved):

PR #152, opened 2023-10-27
AST tests

komalali:


ast/environment.go
```diff
@@ -207,32 +206,7 @@ func (d *EnvironmentDecl) recordSyntax() *syntax.Node {
...
-		syntax:      node,
-		Description: description,
-		Imports:     imports,
-		Values:      values,
-	}
-}
-
-func Environment(description *StringExpr, imports ImportListDecl, values PropertyMapDecl) *EnvironmentDecl {
-	return EnvironmentSyntax(nil, description, imports, values)
-}
```
komalali, line 233: Appears unused. Checked both `pulumi` and `pulumi-service`. But please correct me if I'm wrong!

PR #169, opened 2023-11-06
Add preview tests

komalali:
These tests allow us to more easily catch regressions in environment previews.


lblackstone (approved):
🎉

pgavlin:
Instead of capturing this in a separate file, could we add it to the existing `expected.json`? Also, I'm not certain that these will help catch regressions in previews, as those typically use the result of `Check`, and the values that are being tested here are the result of `Eval`. Can we revert this, make those adjustments, and resubmit?

komalali:
😿

Issue #164, opened 2023-11-03
Editor view always reports unknown for secrets


iwahbe:
### What happened?

I defined an environment with secrets (from `fn::secret`). The preview shows all fields as `[unknown]`, even though they are knowable. Ideally, I would see a fully flushed out object with only secret values redacted. At the very least, the preview should say `[secret]` instead of `[unknown]`.

### Example

https://app.pulumi.com/pulumi/esc/pulumi-cloudflare-testing

### Output of `pulumi about`

NA

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

pgavlin:
@komalali this is a dupe of https://github.com/pulumi/pulumi-service/issues/16162. Would you mind closing that issue as a dupe and using this one to track?

komalali:
Hmm.. that one is the one tracked in the iteration and I don't wanna really change a bunch of things around since I'm pretty close to closing it out. I'll just close them both with the PR.

PR #170, opened 2023-11-06
eval: improve unknown/secret for complex literals 

komalali:
Do not treat object and array literals that include unknown or secret
elements as unknown or secret themselves. Instead, leave it up to the
consumer of the literal to handle unknownness and secretness
appropriately (i.e. by combining the unknownness and secretness of each
leaf value the consumer observes).

For example, the following object definition no is no longer unknown
during check or secret during eval:

```yaml
foo:
  bar: 42
  baz:
    fn::secret:
      ciphertext: AjIAEJFDIe==
```

Instead, it is partially-known during check:

```yaml
foo:
  bar: 42
  baz: [unknown]
```

And partially-secret during eval:

```yaml
foo:
  bar: 42
  baz: [secret]
```

But when passed to `fn::toJSON`, the result of the call would be unknown
in check or secret in eval b/c of the nested unknown/secret.

Fixes https://github.com/pulumi/esc/issues/164.
Fixes https://github.com/pulumi/pulumi-service/issues/16162.

### DIFF
--- a/eval/eval.go
+++ b/eval/eval.go
@@ -494,8 +494,6 @@ func (e *evalContext) evaluateArray(x *expr, repr *arrayExpr) *value {
 	array, items := make([]*value, len(repr.elements)), make([]schema.Builder, len(repr.elements))
 	for i, elem := range repr.elements {
 		ev := e.evaluateExpr(elem)
-		v.combine(ev)
-
 		array[i], items[i] = ev, ev.schema
 	}
 
@@ -516,8 +514,6 @@ func (e *evalContext) evaluateObject(x *expr, repr *objectExpr) *value {
 	object, properties := make(map[string]*value, len(keys)), make(map[string]schema.Builder, len(keys))
 	for _, k := range keys {
 		pv := e.evaluateExpr(repr.properties[k])
-		v.combine(pv)
-
 		object[k], properties[k] = pv, pv.schema
 	}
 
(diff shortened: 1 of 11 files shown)


eval/testdata/eval/schema-error/expected.json
```diff
@@ -255,7 +255,7 @@
         },
         {
             "Severity": 1,
-            "Summary": "dependentReq.bar: missing required property",
```
komalali, line 258: This is odd, investigating

eval/eval_validate.go
```diff
@@ -428,7 +428,7 @@ func (e *validator) validateElement(v *value, accept *schema.Schema, loc validat
 	if accept.Never {
 		return false
 	}
-	if v.unknown {
+	if v.containsUnknowns() {
```
pgavlin, line 431: This one we can leave as-is. The validator understands nested unknowns.

eval/eval.go
```diff
@@ -536,7 +532,7 @@ func (e *evalContext) evaluateInterpolate(x *expr, repr *interpolateExpr) *value
 		if i.value != nil {
 			pv := e.evaluatePropertyAccess(x, i.value.accessors)
 			s, unknown, secret := pv.toString()
-			v.unknown, v.secret = v.unknown || unknown, v.secret || secret
+			v.unknown, v.secret = v.containsUnknowns() || unknown, v.containsSecrets() || secret
```
pgavlin, line 535: I think we need similar changes for `toString` and `toJSON`

pgavlin (changes requested):

eval/eval.go
komalali: `v.combine` is updated to handle this now.

eval/eval_validate.go
komalali: fixed

pgavlin (approved):

PR #172, opened 2023-11-07
Add expected tests for previews

komalali:


environment.go
```diff
@@ -88,3 +88,15 @@ func (e *Environment) GetTemporaryFiles() map[string]Value {
...
+	props := e.Properties
+	json := make(map[string]any, len(props))
+	if props == nil {
+		return nil
+	}
+	for k, v := range props {
+		json[k] = v.ToJSON(true)
+	}
+	return json
+}
```
pgavlin, line 102: Can we remove this method and just call `NewValue(e.Properties).ToJSON(true)`?

eval/eval_test.go
```diff
@@ -218,6 +218,7 @@ func TestEval(t *testing.T) {
 		CheckDiags  syntax.Diagnostics `json:"checkDiags,omitempty"`
 		EvalDiags   syntax.Diagnostics `json:"evalDiags,omitempty"`
 		Environment *esc.Environment   `json:"environment,omitempty"`
+		Preview     map[string]any     `json:"preview,omitempty"`
```
pgavlin, line 221:
```suggestion
		CheckJSON     map[string]any     `json:"checkJson,omitempty"`
```

pgavlin: (it would also be nice to have `EvalJSON` as a parallel, but we can do that as a follow-up)

komalali: fixed and added EvalJSON too

eval/eval_test.go
```diff
@@ -236,18 +238,24 @@ func TestEval(t *testing.T) {
...
+
 				bytes, err := json.MarshalIndent(expectedData{
 					LoadDiags:   loadDiags,
 					CheckDiags:  checkDiags,
 					EvalDiags:   evalDiags,
 					Environment: actual,
+					EvalJSON:    evalJSON,
+					CheckJSON:   checkJSON,
 				}, "", "    ")
+				bytes = append(bytes, '\n')
```
pgavlin, line 258: just curious why we need this bit

pgavlin (approved):

eval/eval_test.go
komalali: coz I hate this icon
Screenshot 2023-11-07 at 1 51 00 PM

PR #176, opened 2023-11-08
Reorder test output for easier diff parsing for humans

komalali:
Reorders the test output so both json outputs are displayed together. Perhaps subjective, but having the two JSON outputs together in the file is a lot easier to read through than having to jump around the larger environment structs.


pgavlin:
Closing this for now.

PR #191, opened 2023-12-05
fromJSON: Mark nested structures as secret if the JSON string is secret

komalali:
The first commit demonstrates the issue, secret objects (and arrays) were not being marked as secret when evaluating fromJSON.

### DIFF
--- a/eval/eval.go
+++ b/eval/eval.go
@@ -919,13 +919,12 @@ func (e *evalContext) evaluateBuiltinFromJSON(x *expr, repr *fromJSONExpr) *valu
 			return v
 		}
 
-		ev, err := esc.FromJSON(jv)
+		ev, err := esc.FromJSON(jv, v.secret)
 		if err != nil {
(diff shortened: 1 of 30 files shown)


eval/eval.go
```diff
@@ -925,6 +925,7 @@ func (e *evalContext) evaluateBuiltinFromJSON(x *expr, repr *fromJSONExpr) *valu
 			v.unknown = true
 			return v
 		}
+		ev.Secret = v.secret
```
pgavlin, line 928: will this be sufficient? I'm worried that this needs to be deeply plumbed through `FromJSON`...

komalali: Good point, fixed now.

eval/testdata/eval/secret-objects/env.yaml
```diff
@@ -0,0 +1,11 @@
+values:
+  secret_object:
+    fn::secret: '{"haha":"business","also":"secret"}'
```
pgavlin, line 3: lol

pgavlin (approved):

Issue #195, opened 2023-12-08
YAML output format for env open


jtmarmon:
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details
When editing the definition, or viewing it by running `env get <envname>`, you get a nice YAML output. Sometimes when debugging you also need to `env open`. It'd be really nice if you could view the output of an `env open` in YAML (both via CLI and the UI), so it's easier to tell if the output is as expected.

PR #204, opened 2023-12-19
Add support for yaml output format

cleverguy25:
Adds support for yaml format in `esc open`. 

Added test to verify.

Fixes [#195](https://github.com/pulumi/esc/issues/195).


pgavlin (approved):

komalali (approved):
Sweet! We will want to do this in the console too! Is there an issue open for that?

PR #205, opened 2023-12-19
Split `esc/` and `esc/cmd` into separate modules.

pgavlin:
These changes peel `esc/cmd` out of the `esc` module s.t. the latter has no dependency on `pulumi/pulumi`.

Unfortunately, the `esc` CLI itself does require a `pulumi/pulumi` dependency so that it can share credential management with the `pulumi` CLI.

These changes should at least simplify the dependency graph for `pulumi/pulumi` somewhat so that `pulumi/pulumi/sdk` can depend on `esc` without circularity.

Unfortunately, there is still circularity between `esc/cmd` and `pulumi/pulumi/pkg`. We could resolve this by splitting off yet _another_ module that contains the CLI implementation but _not_ the default implementations of various bits that refer to `pulumi/pulumi/pkg`. That would probably give us the following modules:

- `github.com/pulumi/esc` <-- no pulumi dependencies
- `github.com/pulumi/esc/cmd` <-- pkg/ and sdk/ dependencies
- `github.com/pulumi/esc/cmd/cli` <-- sdk/ dependency

`pulumi/pulumi/sdk` could then freely depend on `pulumi/esc` and `pulumi/pulumi/pkg` could depend on `github.com/pulumi/esc/cmd/cli`, all without circularity.

The cost, of course, is a more complicated release process for `esc`, wherein the commit/tag flow becomes:

1. tag `esc/v1.2.3`
2. update `esc/cmd/cli` to consume `esc/v1.2.3`, then commit + tag `esc/cmd/cli/v1.2.3`
3. update `esc/cmd` to consume `esc/cmd/cli/v1.2.3`, then commit + tag `esc/cmd/v1.2.3`

That said, that process is _probably_ worth the simplicity of reasoning about dependencies.

Thoughts?


justinvp:
Thanks for digging into this, Pat!

> `pulumi/pulumi/sdk` could then freely depend on `pulumi/esc` and `pulumi/pulumi/pkg` could depend on `github.com/pulumi/esc/cmd/cli`, all without circularity.

This would be ideal. If others find it unpalatable, I hope we can at least consider the changes in this PR so that `pulumi/pulumi/sdk` can depend on `esc` without circularity.

> That said, that process is _probably_ worth the simplicity of reasoning about dependencies.

It's unfortunate that it complicates the release process, but it's likely worth it for the simpler dependencies. We can automate the release process to make it easier on ourselves.

pgavlin:
> Thanks for digging into this, Pat!
> 
> > `pulumi/pulumi/sdk` could then freely depend on `pulumi/esc` and `pulumi/pulumi/pkg` could depend on `github.com/pulumi/esc/cmd/cli`, all without circularity.
> 
> This would be ideal. If others find it unpalatable, I hope we can at least consider the changes in this PR so that `pulumi/pulumi/sdk` can depend on `esc` without circularity.

SG. I'll roll those changes into this PR.

> > That said, that process is _probably_ worth the simplicity of reasoning about dependencies.
> 
> It's unfortunate that it complicates the release process, but it's likely worth it for the simpler dependencies. We can automate the release process to make it easier on ourselves.

Yeah I agree with that.

komalali:
Agreed with what y'all have discussed. Unfortunate about the release process but fixing the circular dependency seems worth it.

PR #213, opened 2024-01-11
eval: fix merging of merged values

### LINKED ISSUE #212: ESC: multiple environments in stack config don't work
### What happened?

I was trying to break up my config into several ESC environments and then import them individually into my Stack config. When more than one environment was set, previous configurations became undefined. This seems to be because one of my environments contained an import. If none of the environments have an import the configuration works as expected but as soon as there is an import then only the environment with the import loads and the others are undefined.

```yaml
# Pulumi.dev.yaml
environment:
  - api-auth-dev
  - api-billing-dev
  - api-documents-dev
config:
  aws:profile: dev
  aws:region: ap-southeast-2
  ...
```

### Example

Environments:
```yaml
# env-a
values:
  pulumiConfig:
    foo: env-a-config
```

```yaml
# env-b
values:
  pulumiConfig:
    bar: env-b-config
```

```yaml
# Pulumi.dev.yaml
environment:
  - env-a
  - env-b
config:
 ...
```

```typescript
// index.ts

const config = new pulumi.Config();

console.log(`config.get("foo"): ${config.get("foo")}`)
console.log(`config.get("bar"): ${config.get("bar")}`)
```
```
Output:
  pulumi:pulumi:Stack (myhr-fullenv-dev):
    config.get("foo"): env-a-config
    config.get("bar"): env-b-config
```

This works as expected - but now if I import something in env-b, env-a will become undefined.

```yaml
# env-b
imports:
  - env-c-secrets
values:
  pulumiConfig:
    bar: env-b-config
```

```yaml
# env-c-secrets
values:
  password: 12345
```

Run the code again this is my output:
```
  pulumi:pulumi:Stack (myhr-fullenv-dev):
    config.get("foo"): undefined
    config.get("bar"): env-b-config
```

### Output of `pulumi about`

```
pulumi about

CLI
Version      3.96.2
Go Version   go1.21.4
Go Compiler  gc

Plugins
NAME     VERSION
aws      6.13.0
command  0.9.2
nodejs   unknown

Host
OS       darwin
Version  14.1
Arch     x86_64

This project is written in nodejs: executable='/Users/thor/.nvm/versions/node/v20.10.0/bin/node' version='v20.10.0'

Current Stack: myhr/myhr-fullenv/dev
```

### Additional context

_No response_

### Contributing

pgavlin:
These changes fix the merging of values that have already been merged.

During evaluation, merged values are represented by _chains_. Each value
`v` maintains a reference to its _base_ value--i.e. the value with which
it has been merged--if any. For example, consider the following values,
where `repr` is the value's data and `base` is the value's base:

```
A: (repr: { foo: B }, base: None)
B: (repr: "bar", base: None)

C: (repr: { foo: D, zed: E }, base: None)
D: (repr: "baz", base: None)
E: (repr: "qux", base: None)
```

When merging `C` onto `A`, `C`'s base will be set to `A` and`D`'s base
will be set to `B`:

```
A: (repr: { foo: B }, base: None)
B: (repr: "bar", base: None)

C: (repr: { foo: D, zed: E }, base: A)
D: (repr: "baz", base: B)
E: (repr: "qux", base: None)
```

Now introduce the following values:

```
F: (repr: { foo: G, zed: H }, base: None)
G: (repr: "beta", base: None)
H: (repr: "gamma", base: None)
```

If `C` is merged onto `F`, `A`'s base should be set to `F`, `B`'s base
should be set to `G`, and `E`'s base should be set to `H`:

```
F: (repr: { foo: G, zed: H }, base: None)
G: (repr: "beta", base: None)
H: (repr: "gamma", base: None)

A: (repr: { foo: B }, base: F)
B: (repr: "bar", base: G)

C: (repr: { foo: D, zed: E }, base: A)
D: (repr: "baz", base: B)
E: (repr: "qux", base: H)
```

In order to achieve this, the merge operation looks like this:

```
def merge(top, base):
  # If there is no new base or if the new base is the same as v, we're
  # done.
  if base == None || top == base:
    return

  if top.base is not None:
    # This value already has a base. Recur.
    merge(top.base, base)
  else:
    # This value does not have a base. Set it now.
    top.base = base

  # Now that the base has been updated, this value's members may need to
  # be updated to point to their new base.
  if isinstance(top.repr, dict):
    for k, p in ctop.repr.items():
      merge(p, top.base.property(k)
```

Prior to these changes, the merge operation had two bugs.

First--and most importantly--it only updated a top value's members if
the top value did not already have a base. Fixing this issue was
straightforward, as it only required adjusting the code to
unconditionally update the top value's members. This is the fix to #212.
In reference to the figures above, prior to these changes, merging `C`
onto `F` would have resulted in:

```
F: (repr: { foo: G, zed: H }, base: None)
G: (repr: "beta", base: None)
H: (repr: "gamma", base: None)

A: (repr: { foo: B }, base: F)
B: (repr: "bar", base: G)

C: (repr: { foo: D, zed: E }, base: A)
D: (repr: "baz", base: B)
E: (repr: "qux", base: None) # <- base should be H
```

The second issue was more subtle. If a value already has a base, merging
that value onto another will mutate the base. This means that base
values cannot be shared between multiple top values, and must instead be
copied. However, the copy routine violated the _critically important
property_ that the base `Q` for a property `P` in a top value must be the
same object as `Q` in the base of the object that contains `P`. For
example, given:

```
A: (repr: { prop: Q }, base: None)
                  ^ call this reference R1

Q: (repr: "foo", base: None)

B: (repr: { prop: P }, base: A)
P: (repr: "bar", base: Q)
                       ^ call this reference R2
```

`R1` and `R2` must refer to the same `Q`. If they do not, then merging
`A` and `B` will result in redundant merges. This isn't a correctness
issue as far as the resulting value, but it does cause incorrect traces.

In order to fix this issue, these changes replace the `(*value).copy`
method with `(copier).copy`, which tracks copied values and ensures that
all references to a value `V` are redirected to the same copy of `V`
rather than copying `V` multiple times.

Fixes #212.

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -2,3 +2,5 @@
 
 ### Bug Fixes
 
+- Fix merging of already-merged values.
+  [#213](https://github.com/pulumi/esc/pull/213)
--- a/eval/eval.go
+++ b/eval/eval.go
@@ -317,6 +317,7 @@ func (e *evalContext) evaluate() (*value, syntax.Diagnostics) {
 	// root.
 	properties := make(map[string]*expr, len(e.env.Values.GetEntries()))
 	e.root = &expr{
+		path: fmt.Sprintf("<%v>", e.name),
 		repr: &objectExpr{
 			node:       ast.Object(),
 			properties: properties,
@@ -414,7 +415,7 @@ func (e *evalContext) evaluateImport(myImports map[string]*value, decl *ast.Impo
 
 	myImports[name] = val
 	if merge {
-		val = val.copy()
+		val = newCopier().copy(val)
 		val.merge(e.base)
 		e.base = val
 	}
@@ -566,7 +567,7 @@ func (e *evalContext) evaluateInterpolate(x *expr, repr *interpolateExpr) *value
 func (e *evalContext) evaluatePropertyAccess(x *expr, accessors []*propertyAccessor) *value {
 	// We make a copy of the resolved value here because evaluateExpr will merge it with its base, which mutates the
 	// value. We also stamp over the def with the provided expression in order to maintain proper error reporting.
-	v := e.evaluateExprAccess(x, accessors).copy()
+	v := newCopier().copy(e.evaluateExprAccess(x, accessors))
 	v.def = x
 	return v
 }
--- a/eval/eval_test.go
+++ b/eval/eval_test.go
@@ -276,11 +276,19 @@ func TestEval(t *testing.T) {
 				assert.Equal(t, expected.EvalJSONRedacted, evalJSONRedacted)
 				evalJSONRevealed := esc.NewValue(actual.Properties).ToJSON(false)
 				assert.Equal(t, expected.EvalJSONRevealed, evalJSONRevealed)
+
+				bytes, err := json.MarshalIndent(evalJSONRevealed, "", "  ")
+				require.NoError(t, err)
+				t.Logf("eval: %v", string(bytes))
 			}
 
 			if check != nil {
 				checkJSON := esc.NewValue(check.Properties).ToJSON(true)
 				assert.Equal(t, expected.CheckJSON, checkJSON)
+
+				bytes, err := json.MarshalIndent(checkJSON, "", "  ")
+				require.NoError(t, err)
+				t.Logf("check: %v", string(bytes))
 			}
 
 			assert.Equal(t, expected.Check, check)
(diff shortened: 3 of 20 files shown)


pgavlin:
Working on adding some additional tests here. This is super subtle.

komalali (approved):
Thank you for the very helpful PR description!

cleverguy25 (approved):

pgavlin:
> Thank you for the very helpful PR description!

No problem. I'm glad it helped!

PR #239, opened 2024-02-02
Add support for execution context data for interpolation

glena:
Add support for injecting contextual execution information and enable it to be used to interpolate in the environment attributes.

IE:
```
context = {
     "attr1": { "value": "foo" }, 
     "attr2": "bar"
}
```
can be used as:

```
values:
    sample: ${context.attr1.value}-${context.attr2}
```

would generate:

```
{
    "sample": "foo-bar"
}
```

### DIFF
--- a/eval/eval.go
+++ b/eval/eval.go
@@ -336,6 +336,7 @@ func (e *evalContext) isReserveTopLevelKey(k string) bool {
 func (e *evalContext) evaluate() (*value, syntax.Diagnostics) {
 	// Evaluate imports. We do this prior to declaration so that we can plumb base values as part of declaration.
 	e.evaluateImports()
+	// Evaliate context. We prepare the context values to later evaluate interpolations.
 	e.evaluateContext()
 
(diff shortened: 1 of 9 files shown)


cleverguy25 (approved):

eval/eval.go
```diff
@@ -77,8 +77,9 @@ func EvalEnvironment(
 	decrypter Decrypter,
 	providers ProviderLoader,
 	environments EnvironmentLoader,
+	context esc.Value,
```
pgavlin, line 80: how about we make this a `map[string]esc.Value`? we could always relax it later if we need to, but I think that'll simplify callers of this function a bit and it's sufficient for our needs.

pgavlin (approved):
one suggestion to simplify callers, but LGTM otherwise!

eval/eval.go
glena: Done, I end up wrapping the `map[string]esc.Value` into a `Value` internally to keep it simple (otherwise I would have to write a new `unexport` to handle that case and/or change how `evaluateExprAccess` process the value. I believe this makes it more straightforward.

PR #248, opened 2024-02-06
eval: add a test for a missing import

pgavlin:
Just what it says on the tin. Ensures that evaluation does not panic in the case of a missing import.


pgavlin:
(confirmed that the new test fails w/o the recent changes we made to avoid panics)

lblackstone (approved):

cleverguy25 (approved):

PR #250, opened 2024-02-09
Handle anonymous environments when injecting it to the context

glena:
- Splitting `context.environment` into `rootEnvironment` and `currentEnvironment` (potentially just `environment`)
- When evaluating an anonymous environment, the `rootEnvironment` will evaluate to the rootest non anonymous environment for the imported environments (wont propagate to the root)

Having these 3 envs:

```
# root
imports:
  - imported
values:
  root:
    currentEnvironment: ${context.currentEnvironment.name}
    rootEnvironment: ${context.rootEnvironment.name}

# imported
imports:
  - child
values:
  imported:
    currentEnvironment: ${context.currentEnvironment.name}
    rootEnvironment: ${context.rootEnvironment.name}

# child
values:
  child:
    currentEnvironment: ${context.currentEnvironment.name}
    rootEnvironment: ${context.rootEnvironment.name}
``` 

In the case the `root` env is anonymous, it resolves:
```
{
  "child": {
    "currentEnvironment": "child",
    "rootEnvironment": "imported"
  },
  "imported": {
    "currentEnvironment": "imported",
    "rootEnvironment": "imported"
  },
  "root": {
    "currentEnvironment": "<yaml>",
    "rootEnvironment": "<yaml>"
  }
}
```

In the case the `root` env is a named environment, it resolves:
```
{
  "child": {
    "currentEnvironment": "child",
    "rootEnvironment": "root"
  },
  "imported": {
    "currentEnvironment": "imported",
    "rootEnvironment": "root"
  },
  "root": {
    "currentEnvironment": "root",
    "rootEnvironment": "root"
  }
}
```

### DIFF
--- a/environment.go
+++ b/environment.go
@@ -29,10 +29,55 @@ type ExecContext struct {
 	values         map[string]Value
 }
 
+type copier struct {
+	memo map[*Value]*Value
+}
+
+func newCopier() copier {
+	return copier{memo: map[*Value]*Value{}}
+}
+
+func (c copier) copy(v *Value) *Value {
+	if v == nil {
+		return nil
+	}
+
+	if copy, ok := c.memo[v]; ok {
+		return copy
+	}
+
+	copy := &Value{}
+	c.memo[v] = copy
+
+	var nv any
+	switch vr := v.Value.(type) {
+	case []*Value:
+		a := make([]*Value, len(vr))
+		for i, v := range vr {
+			a[i] = c.copy(v)
+		}
+		nv = a
+	case map[string]*Value:
+		m := make(map[string]*Value, len(vr))
+		for k, v := range vr {
+			m[k] = c.copy(v)
+		}
+		nv = m
+	default:
+		nv = vr
+	}
+
+	*copy = Value{
+		Value: nv,
+	}
+	return copy
+}
+
 func copyContext(context map[string]Value) map[string]Value {
 	newContext := make(map[string]Value)
 	for key, value := range context {
-		newContext[key] = value
+		copy := newCopier().copy(&value)
+		newContext[key] = *copy
 	}
 	return newContext
 }
--- a/environment.go
+++ b/environment.go
@@ -85,13 +85,13 @@ func copyContext(context map[string]Value) map[string]Value {
 
 func (ec *ExecContext) Copy() *ExecContext {
 	return &ExecContext{
-		values:         copyContext(ec.values),
+		values:         ec.values,
 		rootEvironment: ec.rootEvironment,
 	}
 }
 
 func (ec *ExecContext) Values(envName string) map[string]Value {
-	context := ec.values
+	context := copyContext(ec.values)
 	context["currentEnvironment"] = NewValue(map[string]Value{
 		"name": NewValue(envName),
 	})
@@ -107,11 +107,11 @@ func (ec *ExecContext) Values(envName string) map[string]Value {
 	return context
 }
 
-var ErrForbiddenContextkey = errors.New("forbidden context key")
+var ErrReservedContextkey = errors.New("reserved context key")
 
 func validateContextVariable(context map[string]Value, key string) error {
 	if _, ok := context[key]; ok {
-		return fmt.Errorf("%w: %q", ErrForbiddenContextkey, key)
+		return fmt.Errorf("%w: %q", ErrReservedContextkey, key)
 	}
 	return nil
 }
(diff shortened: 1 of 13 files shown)


environment.go
```diff
@@ -16,10 +16,74 @@ package esc
...
 )
 
+const AnonymousEnvironmentName = "<yaml>"
+
+type ExecContext struct {
+	rootEvironment string
+	values         map[string]Value
+}
+
+func copyContext(context map[string]Value) map[string]Value {
```
glena, line 32: It is still pending change this to a deep copy instead of a shallow one

glena: I will look into unifying it with eval's copier to have a single implementation

environment.go
```diff
@@ -16,10 +16,120 @@ package esc
...
+	rootEvironment string
+	values         map[string]Value
+}
+
+type copier struct {
+	memo map[*Value]*Value
+}
+
+func newCopier() copier {
+	return copier{memo: map[*Value]*Value{}}
```
glena, line 37: I can merge this to reuse the copier between eval/val and this

https://github.com/glena/esc/compare/glena:esc:german/root-current-env-context...glena:esc:german/root-current-env-context-copy?expand=1

environment.go
```diff
@@ -16,10 +16,120 @@ package esc
...
+	if ec.rootEvironment == AnonymousEnvironmentName || ec.rootEvironment == "" {
+		ec.rootEvironment = envName
+	}
+
+	context["rootEnvironment"] = NewValue(map[string]Value{
+		"name": NewValue(ec.rootEvironment),
+	})
+
+	return context
+}
```
pgavlin, line 108: I'd prefer not to mutate the values here. I'd prefer that we instead copy the context for each `envName` and set the values in `Copy`.

environment.go
```diff
@@ -16,10 +16,120 @@ package esc
...
+
+var ErrForbiddenContextkey = errors.New("forbidden context key")
```
pgavlin, line 110: nit: we generally use "reserved" for keys that the user is not allowed to set

```suggestion
var ErrReservedContextkey = errors.New("reserved context key")
```

environment.go
pgavlin: I think it's fine to have two implementations for now: using the generic copier feels a bit risky to me given the sensitivity of the `*value` graph to proper memoized copies.

FWIW, the `*Value`s don't have the same restrictions, so we could also simplify this by removing the memoization.

pgavlin (changes requested):
one change re: not mutating execcontexts, but generally looks good.

environment.go
glena: done

environment.go
glena: done

environment.go
```diff
@@ -16,10 +16,120 @@ package esc
...
+	if ec.rootEvironment == AnonymousEnvironmentName || ec.rootEvironment == "" {
+		ec.rootEvironment = envName
+	}
+
+	context["rootEnvironment"] = NewValue(map[string]Value{
+		"name": NewValue(ec.rootEvironment),
+	})
+
+	return context
+}
```
pgavlin, line 108: sorry, I think I wasn't clear here. this is what I was suggesting:

```suggestion
func (ec *ExecContext) Copy(envName string) *ExecContext {
	values := copyContext(ec.values)
	values["currentEnvironment"] = NewValue(map[string]Value{
		"name": NewValue(envName),
	})

	root := ec.rootEnvironment
	if ec.rootEvironment == AnonymousEnvironmentName || ec.rootEvironment == "" {
		root = envName
	}

	values["rootEnvironment"] = NewValue(map[string]Value{
		"name": NewValue(root),
	})

	return &ExecContext{
		values:         values,
		rootEvironment: root,
	}
}

func (ec *ExecContext) Values() map[string]Value {
	return ec.values
}
```

With this approach the contents of a context are never mutated once the context is created.

pgavlin (changes requested):

environment.go
glena: ah I misunderstood. I thought the concern was mutating the value we have in memory, not the place were we mutate. Done

environment.go
```diff
@@ -16,10 +16,120 @@ package esc
...
+
+func (ec *ExecContext) CopyForEnv(envName string) *ExecContext {
+	context := copyContext(ec.values)
+	context["currentEnvironment"] = NewValue(map[string]Value{
+		"name": NewValue(envName),
+	})
+
+	if ec.rootEvironment == AnonymousEnvironmentName || ec.rootEvironment == "" {
+		ec.rootEvironment = envName
+	}
```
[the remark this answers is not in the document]
glena: yes that is needed to propagate the root environment down the line

pgavlin (approved):

PR #255, opened 2024-02-15
Fix anonymous env test

glena:
Having characters such as `<` or `>` is not allowed by go to keep compatibility with different environments. To fix this, I am renaming the directory and adding a new `overrides` file to be able to not follow the conventions in cases like this.


eval/eval_test.go
```diff
@@ -211,16 +230,34 @@ func TestEval(t *testing.T) {
...
+				require.NoError(t, err)
+				dec := json.NewDecoder(bytes.NewReader(expectedBytes))
+				dec.UseNumber()
+				err = dec.Decode(&overrides)
+				require.NoError(t, err)
+
+				if overrides.RootEnvironment != "" {
+					environmentName = overrides.RootEnvironment
+				}
+			}
```
pgavlin, line 249: can probably simplify this a bit:

```suggestion
			environmentName := e.Name()
			overridesBytes, err := os.ReadFile(overridesPath)
			require.True(t, err == nil || errors.Is(err, os.ErrNotExist))
			if err == nil {
				var overrides testOverrides
				err = json.Unmarshal(overridesBytes, &overrides)
				require.NoError(t, err)
				environmentName = overrides.RootEnvironment
			}
```

then we can remove the `doesFileExist` function as well.

pgavlin (commented):
one suggestion for simplifying the code. LGTM otherwise

eval/eval_test.go
glena: makes sense

pgavlin (approved):

PR #259, opened 2024-02-16
chore: freeze v0.8.2

pgavlin:


.version
```diff
@@ -1 +1 @@
-0.8.1
+0.8.2
```
cleverguy25, line 1: PR named wrong? Shouldn't be v0.8.2?

pgavlin: fixed

cleverguy25 (approved):

PR #263, opened 2024-02-28
Specify pulumi access token per command run

glena:


cmd/esc/cli/login.go
```diff
@@ -157,7 +154,28 @@ func (esc *escCommand) getCachedClient(ctx context.Context) error {
...
+			display.Options{Color: esc.colors},
+		)
+
+		if err != nil {
+			return fmt.Errorf("could not determine current cloud: %w", err)
+		}
+
+		account = &workspace.Account{
+			Account:    *nAccount,
+			BackendURL: "https://api.pulumi.com",
```
pgavlin, line 176:
```suggestion
			BackendURL: backendURL,
```

cmd/esc/cli/login.go
```diff
@@ -157,7 +154,28 @@ func (esc *escCommand) getCachedClient(ctx context.Context) error {
...
+			DefaultOrg: "",
```
pgavlin, line 177: we probably want to populate the default org here as well (see `workspace.GetBackendConfigDefaultOrg`)

cmd/esc/cli/login.go
glena: 🤦

cmd/esc/cli/login.go
glena: done

cmd/esc/cli/login.go
```diff
@@ -157,7 +154,33 @@ func (esc *escCommand) getCachedClient(ctx context.Context) error {
...
+		nAccount, err := esc.login.Login(
+			ctx,
+			backendURL,
+			false,
+			"esc",
+			"Pulumi ESC environments",
+			nil,
+			false,
+			display.Options{Color: esc.colors},
+		)
```
pgavlin, line 168: I think the only remaining question I have is what the experience is like if no access token is specified via `PULUMI_ACCESS_TOKEN`? If we fall through to the `err != nil` case below I think that we'll be regressing the UX, as the user will see `could not determine current cloud` instead of `no credentials. Please run esc login to log in.`

glena: It is the same experience you have with the pulumi cli where it asks you to authenticate

image

image

pgavlin: excellent; thanks!

pgavlin: what about in non-interactive scenarios?

pgavlin (approved):

cmd/esc/cli/login.go
pgavlin: (try running with the output redirected, for example)

glena: This is the output
image

PR #271, opened 2024-03-26
Initial golang and typescript SDK and templates

cleverguy25:
This is a large PR with lots of generated files, so here is a walkthrough.  Basically each language SDK has templates, those templates are used to generate a "raw" API, and there is a handcrafted but light wrapper to "cleanup" the result or provide convenience methods.

- ./.vscode/
  - settings to run/debug unit tests in vscode
- ./Makefile
  - add code generation commands
- ./sdk/esc-api.rest
  - vscode REST extension file to play with APIs
- ./sdk/templates
  - downloaded templates using `openapi-generator-cli author template -g <generator> -o -./sdk/templates/<generator_output_dest>
  - This is so we have control but also for go and python we need to add yaml support and some files to support a "raw" client
- ./sdk/swagger.yaml
  - handcrafted swagger/openapi spec
- ./sdk/**/.openapi-generator-ignore
  - suppresses generation of superfluous files
- ./sdk/go/api_esc_extensions.go
  - handcrafted wrapper over the raw API client
- ./sdk/go/api_esc_test.go
  - handcrafted E2E tests
- ./sdk/typescript/esc/index.ts
  - handcrafted wrapper over the raw API client
- ./sdk/typescript/test/client.spec.ts
  - hand crafted E2E tests


pgavlin (commented):
A few questions here:

- How will we keep the SDK up to date?
- Is the generated code deterministic?
- Can we add a target for testing the TS/JS SDK (or is that already covered?)
- Can we add a step to our GH action to invoke the code generator + run tests?

cleverguy25:
> A few questions here:
> 
> * How will we keep the SDK up to date?

Code gen does a lot of the heavy lifting, but the wrapper requires manual work.  But that really took me a couple hours and once boilerplate there was easier as it went along.

> * Is the generated code deterministic?

Because we download  the templates it is very deterministic.

> * Can we add a target for testing the TS/JS SDK (or is that already covered?)

What do you mean by target?  Like code coverage?  Right now the code coverage is low (like 33% last I checked) because the code generated models are very detailed but I think the prime scenarios are well covered.

> * Can we add a step to our GH action to invoke the code generator + run tests?

We can certainly run tests.  I would prefer, at least initially, that we choose when we run the generator as it really only matters when we update the swagger.yaml doc, but even then would want the dev to run it and include it in the PR.

pgavlin:
> > A few questions here:
> > 
> > * How will we keep the SDK up to date?
> 
> Code gen does a lot of the heavy lifting, but the wrapper requires manual work. But that really took me a couple hours and once boilerplate there was easier as it went along.

👍 

> 
> > * Is the generated code deterministic?
> 
> Because we download the templates it is very deterministic.

👍

> 
> > * Can we add a target for testing the TS/JS SDK (or is that already covered?)
> 
> What do you mean by target? Like code coverage? Right now the code coverage is low (like 33% last I checked) because the code generated models are very detailed but I think the prime scenarios are well covered.

A target in the `Makefile` that we can run (or just include this in the `test` target).

> > * Can we add a step to our GH action to invoke the code generator + run tests?
> 
> We can certainly run tests. I would prefer, at least initially, that we choose when we run the generator as it really only matters when we update the swagger.yaml doc, but even then would want the dev to run it and include it in the PR.

This is contrary to our historical approach to code generation. In the past we've run codegen with each PR and then checked for a dirty working tree in order to ensure that our generated code is out-of-date. I have a slight preference for this style, as it makes it easier to ensure that the swagger definition doesn't drift from the generated SDK.

sdk/swagger.yaml
```diff
@@ -0,0 +1,720 @@
...
+        list:
+          type: array
+          items:
+            $ref: "#/components/schemas/Expr"
+        object:
+          type: object
+          additionalProperties:
+            $ref: "#/components/schemas/Expr"
+        builtin:
+          $ref: "#/components/schemas/ExprBuiltin"
```
pgavlin, line 137: This really ought to be a `oneOf`--only one of these properties can ever be present at a time. Do `oneOf`s play well with the various code generators?

sdk/swagger.yaml
```diff
@@ -0,0 +1,720 @@
...
+      type: object
+      properties:
+        def: 
+          $ref: "#/components/schemas/Range"
+        base:
+          $ref: "#/components/schemas/Value"
+    Value:
+      type: object
+      properties:
+        value: {}
```
pgavlin, line 175: We could tighten this up, I think, by making it a `oneOf`. The type is `null | boolean | number | string | Value[] | Record<string, Value>`. Not sure how that would play with the code generators, though.

sdk/swagger.yaml
```diff
@@ -0,0 +1,720 @@
...
+    EnvironmentDefinitionValues:
+      type: object
+      properties:
+        pulumiConfig:
+          type: object
+          additionalProperties: true
+        environmentVariables:
+          type: object
+          additionalProperties:
+            type: string
```
pgavlin, line 230: Why are these present? If we're going to have them, we should also have `files`.

sdk/templates/go/README.mustache
```diff
@@ -0,0 +1,235 @@
+# Go API client for {{packageName}}
```
pgavlin, line 1: You mentioned that these templates are downloaded. Out of curiosity, where do they come from? If we needed to update them, how would we do so?

sdk/go/api_esc_extensions.go
```diff
@@ -0,0 +1,287 @@
+package esc
```
pgavlin, line 1: Nit: this package name will conflict with the name of the root `esc` package. That might be fine--just wanted to call it out.

sdk/go/api_esc_extensions.go
```diff
@@ -0,0 +1,287 @@
...
+
+		return output
+	} else if sliceData, isSlice := value.([]any); isSlice {
+		for i, v := range sliceData {
+			sliceData[i] = mapValuesPrimitive(v)
+		}
+		return sliceData
+	}
+
+	return value
```
pgavlin, line 172: More idiomatic to express this as a type switch:

```suggestion
	switch val := value.(type) {
	case *Value:
		return mapValuesPrimitive(val.Value)
	case map[string]Value:
		output := make(map[string]any, len(mapData))
		for k, v := range mapData {
			output[k] = mapValuesPrimitive(v.Value)
		}
		return output
	case []any:
		for i, v := range sliceData {
			sliceData[i] = mapValuesPrimitive(v)
		}
		return sliceData
	default:
		return value
	}
```

sdk/go/api_esc_extensions.go
```diff
@@ -0,0 +1,287 @@
...
+func getMapSafe(data any) map[string]any {
+	if data != nil {
+		if val, ok := data.(map[string]any); ok {
+			return val
+		}
+
+		return nil
+	}
+
+	return nil
```
pgavlin, line 244: This can be simplified:

```suggestion
	val, _ := data.(map[string]any)
	return val
```

sdk/go/api_esc_extensions.go
```diff
@@ -0,0 +1,287 @@
...
+		for i, v := range sliceData {
+			sliceData[i] = mapValues(v)
+		}
+		return sliceData
+	}
+
+	return value
+}
+
+func getValue(data map[string]any) *Value {
```
pgavlin, line 207: Looks like we're dealing with a lot of `any`s in this code. Is that the result of challenges with the code generator? Can we get more strongly-typed values to work with?

sdk/go/api_esc_extensions.go
```diff
@@ -0,0 +1,287 @@
...
+	env, _, err := c.EscAPI.ReadOpenEnvironment(ctx, org, envName, openEnvID).Execute()
+	if err != nil {
+		return nil, nil, err
+	}
+
+	propertyMap := *env.Properties
+	for k, v := range propertyMap {
+		v.Value = mapValues(v.Value)
+		propertyMap[k] = v
+	}
```
pgavlin, line 61: This is mutating the property map in the environment. Is that intentional?

sdk/go/api_esc_extensions.go
```diff
@@ -0,0 +1,287 @@
...
+func (c *EscClient) GetEnvironment(ctx context.Context, org, envName string) (*EnvironmentDefinition, string, error) {
+	env, resp, err := c.EscAPI.GetEnvironment(ctx, org, envName).Execute()
+	if err != nil {
+		return nil, "", err
+	}
+
+	body, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return nil, "", err
+	}
```
pgavlin, line 41: What is the contents of `body` here vs. the contents of `env`? Is `body` the raw text?

Also, I notice that `EnvironmentDefinition` only contains `json` struct tags--it seems unlikely to unmarshal properly given that most definitions will not be valid JSON. Is that a code generator issue?

sdk/swagger.yaml
cleverguy25: Sometimes, it has been hit or miss.  But a lot of this I pulled from the schema swagger doc, I will have to check if I missed a oneOf there.

sdk/swagger.yaml
cleverguy25: The value generators can do one of, but if they have references to other types or circular references they are angry

sdk/swagger.yaml
cleverguy25: I wondered if you would push back on that :-).  I feel they make the API easier to use, but am not married to them.  Happy to add files.

sdk/templates/go/README.mustache
cleverguy25: The command line outputs them `openapi-generator-cli author template -g typescript-axios -o ./typscript `

sdk/go/api_esc_extensions.go
cleverguy25: The issue is that generator but also swagger doesn't support that as it creates circular references.

sdk/go/api_esc_extensions.go
cleverguy25: Yes, in order to make it more strongly typed.  The first level is of type Value (strongly typed).  This iterates and makes child values type Value too (they start as any)

sdk/go/api_esc_extensions.go
cleverguy25: Yes, body in case people just want to work with the text (run regex or grep or something).

So far I have not seen an issue with the marshalling, but it is possible I am not taxing it enough.  Any example you can think of to test that?  This is why there is a raw Yaml version as well.

sdk/swagger.yaml
cleverguy25: I tried and this cannot do the references to other types because it creates circular references, and the generators will fail.

sdk/swagger.yaml
cleverguy25: The isssue with oneOf here is that it expects that to consolidate to one property in the serialized result, like a Value property that can be oneOf several different schemas.  The generator doesn't have support how we are serializing the output here, and really doesn't have great discriminated union support in general.

komalali (approved):

PR #275, opened 2024-03-27
cli: add revision/tag support to get and open

pgavlin:
These changes add support to the CLI for getting or opening environments at specific revisions or tags using the familiar `name:revision-or-tag` syntax.


pgavlin:
Follow-up changes will add CRUDL support for tags and support for listing environment revisions.

cleverguy25 (approved):

PR #277, opened 2024-03-28
cli: esc env log

pgavlin:
These changes add a command to the CLI that displays the revision history of an environment. This command, `esc env log`, is modeled after `git log`. If a specific version of an environment is given, the history begins at the corresponding revision.

The most complicated part of these changes is the pager support. If a system pager is available, the output is sent to that pager. Like `git`, the pager defaults to `less`, but can be changed via the `--pager` flag or the `PAGER` environment variables. If no pager is available, however, we use a built-in pager. This is primarily intended to address platforms where `less` or other pagers are not present (e.g. stock Windows).


pgavlin:
These changes include a few fixes to the revision support in the CLI test framework.

cleverguy25 (approved):

PR #283, opened 2024-04-10
cli: add commands for managing version tags

pgavlin:
These changes add commands for managing version tags:

- `esc env version tag` creates, describes, or updates a version tag
- `esc env version rm` deletes a version tag
- `esc env version ls` lists version tags

### DIFF
--- a/cmd/esc/cli/cli_test.go
+++ b/cmd/esc/cli/cli_test.go
@@ -11,6 +11,7 @@ import (
 	"hash/fnv"
 	"io"
 	"io/fs"
+	"net/http"
 	"os"
 	"os/exec"
 	"path"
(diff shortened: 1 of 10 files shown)


pgavlin:
w.r.t. the location of the commands: I kind of also like to alias these to `esc version`, but we're currently using that command to print the version of the `esc` CLI. If we want to add that alias, then we could move the "print version" command to `esc about version` or similar.

cmd/esc/cli/cli_test.go
```diff
@@ -635,6 +639,122 @@ func (c *testPulumiClient) ListEnvironmentRevisions(
...
+
+	rev := len(env.revisions)
+	if revision != nil {
+		rev = *revision
+	}
+	if rev < 1 || rev > len(env.revisions) {
+		return errors.New("not found")
+	}
+
+	if _, ok := env.tags[tagName]; !ok {
```
cleverguy25, line 711: Should this just create it in this case?

pgavlin: No--that would differ from the behavior of the ESC API.

pgavlin: (this code is the test-only implementation of the REST client, which fakes up the ESC API)

cleverguy25: Should we consider changing the ESC API to handle an upsert case?

pgavlin: Worth considering, yeah. It'd save us an API call and improve the clarity a bit.

pgavlin: (I'm going to leave it out of scope for this PR, though)

cleverguy25 (approved):

pgavlin:
> w.r.t. the location of the commands: I kind of also like to alias these to `esc version`, but we're currently using that command to print the version of the `esc` CLI. If we want to add that alias, then we could move the "print version" command to `esc about version` or similar.

Any thoughts on this? Strictly speaking a change to `esc version` is a breaking change, but I doubt that it will have any practical effect.

Might also want to move `esc env log` to `esc env version log`.

PR #291, opened 2024-04-18
cli: add tags to `env log` output

pgavlin:
If a revision has tags, display them after the revision number ala `git log`.

### DIFF
--- a/cmd/esc/cli/testdata/env-log.yaml
+++ b/cmd/esc/cli/testdata/env-log.yaml
@@ -40,7 +40,7 @@ stdout: |+
   Author: test-tester <Test Tester>
   Date:   1970-01-01 04:00:00 +0000 UTC
 
-  revision 3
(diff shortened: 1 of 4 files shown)


cmd/esc/cli/testdata/env-log.yaml
```diff
@@ -40,7 +40,7 @@ stdout: |+
   Author: test-tester <Test Tester>
   Date:   1970-01-01 04:00:00 +0000 UTC
 
-  revision 3
+  revision 3 (tag: stable)
```
komalali, line 43: Would be nice to have a test case with multiple tags

komalali (approved):

cmd/esc/cli/testdata/env-log.yaml
pgavlin: Agree--will do that as a follow-up. It's a bit involved.

PR #292, opened 2024-04-18
cli: make `env log` prettier

pgavlin:
Use the same styling as is used for rendering top-level headers in the output of `esc env get`.


pgavlin:
Before: https://asciinema.org/a/hZv2yADyDDXyDOeL33aockpwU

After: https://asciinema.org/a/mErwR5uttPOrU83O2obmWkTw9

komalali (approved):

PR #295, opened 2024-04-30
cli: add a command for diffing revisions

pgavlin:
These changes add a command, `esc env diff`, for displaying the differences between environment revisions. By default, the output includes both value and definition diffs. Diffs between the value as rendered in different formats (e.g. `shell`, `json`, etc.) can also be displayed.

### DIFF
--- a/cmd/esc/cli/env_diff.go
+++ b/cmd/esc/cli/env_diff.go
@@ -1,0 +1,137 @@
+// Copyright 2023, Pulumi Corporation.
+
+package cli
+
+import (
+	"bytes"
+	"context"
+	"fmt"
+
+	"github.com/charmbracelet/glamour"
+	"github.com/spf13/cobra"
+
+	"github.com/pulumi/esc/cmd/esc/cli/style"
+	"github.com/pulumi/pulumi/sdk/v3/go/common/resource"
+	"github.com/pulumi/pulumi/sdk/v3/go/common/util/cmdutil"
+)
+
+func newEnvDiffCmd(env *envCommand) *cobra.Command {
+	var format string
+	var showSecrets bool
+	var pathString string
+
+	diff := &envGetCommand{env: env}
+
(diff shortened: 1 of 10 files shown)


pgavlin:
We might want to send the output of this command through the pager (ala `git diff`). Thoughts?

cmd/esc/cli/env_diff.go
```diff
@@ -0,0 +1,137 @@
...
+	var format string
+	var showSecrets bool
+	var pathString string
+
+	diff := &envGetCommand{env: env}
+
+	cmd := &cobra.Command{
+		Use:   "diff [<org-name>/]<environment-name>[:<revision-or-tag>] [<revision-or-tag>]",
+		Args:  cobra.RangeArgs(1, 2),
+		Short: "Show changes between revisions.",
```
cleverguy25, line 28: Should this say versions since it support revisions and tags

cmd/esc/cli/env_get.go
```diff
@@ -155,7 +153,86 @@ func (get *envGetCommand) showValue(
...
+	var base strings.Builder
+	if err := get.writeValue(ctx, &base, orgName, envName, baseRevisionOrTag, path, format, showSecrets); err != nil {
+		return err
+	}
+	var tip strings.Builder
+	if err := get.writeValue(ctx, &tip, orgName, envName, tipRevisionOrTag, path, format, showSecrets); err != nil {
+		return err
+	}
+
+	baseRef := fmt.Sprintf("%s:%s", envName, baseRevisionOrTag)
```
cleverguy25, line 206: Did we land on @ instead? Assume that would be a separate PR

cleverguy25 (approved):
Some nits

cmd/esc/cli/env_get.go
pgavlin: We did--that'll be a separate PR

PR #306, opened 2024-05-17
cli: prefer '@version' when env name is implied

pgavlin:
`env diff` and `env version tag` both accept arguments that imply an environment name. In the case of `env diff`, this is the version to use for comparison; for `env version tag` it is the version to target with the tag. In both cases, document that the syntax for these arguments is `@version` (but continue to accept a bare version).


pgavlin:
I have follow-up changes that will allow `env diff` to compare different environments, but they're a bit more involved.

lblackstone (approved):

PR #317, opened 2024-05-29
cli: move to new versioning API routes

pgavlin:
All versioning-related APIs now live under `/versions`, and the shape of the revision-related APIs has changed somewhat.

This has the nice side-effect of removing an API call when using tagged versions.


IaroslavTitov (approved):
Unrelated to the PR, but just curious: any reason why these paths use %v and not %s? Afaik, %v is most generic, wouldn't it be safer to use %s?

pgavlin:
> Unrelated to the PR, but just curious: any reason why these paths use %v and not %s? Afaik, %v is most generic, wouldn't it be safer to use %s?

`%s` would let the linters catch cases where we're passing non-strings, yes. But in these cases we also don't really care--if we were writing in TS we'd probably just use string interpolations for these cases, which is morally what we're doing with `%v`.

PR #327, opened 2024-06-03
cli: buffer output to system pager

pgavlin:
Experimentally, this helps avoid problems with corrupted output when piping to the system pager.

These changes also disable the `bubbletea`-based fallback pager, instead falling back to no paging. The `bubbletea` pager just doesn't seem to work properly, and we need a quick fix here. Hopefully we can get a better root cause for all of this in the future.


cmd/esc/cli/pager/noop.go
```diff
@@ -0,0 +1,32 @@
+// Copyright 2023, Pulumi Corporation.
```
dschaller, line 1:
```suggestion
// Copyright 2024, Pulumi Corporation.
```

dschaller (approved):

PR #330, opened 2024-06-04
cli: add env version retract

pgavlin:
This command retracts a specific revision of an environment. A retracted revision can no longer be read or opened. Retracting a revision also updates any tags that point to the retracted revision to instead point to a replacement revision. If no replacement is specified, the latest non-retracted revision preceding the revision being retracted is used as the replacement.

The revision pointed to by the `latest` tag may not be retracted. To retract the latest revision of an environment, first update the environment with a new definition.

### DIFF
--- a/cmd/esc/cli/env_version_retract.go
+++ b/cmd/esc/cli/env_version_retract.go
@@ -43,6 +43,7 @@ func newEnvVersionRetractCmd(env *envCommand) *cobra.Command {
 			if ref.version == "latest" {
 				return errors.New("cannot retract the `latest` revision")
 			}
+			_ = args
 
 			var replacementRevision *int
 			if replacement != "" {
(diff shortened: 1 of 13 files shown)


cmd/esc/cli/env_version_retract.go
```diff
@@ -0,0 +1,70 @@
...
+
+	"github.com/spf13/cobra"
+)
+
+func newEnvVersionRetractCmd(env *envCommand) *cobra.Command {
+	var replacement string
+	var reason string
+
+	cmd := &cobra.Command{
+		Use:   "retract [<org-name>/]<environment-name>@<version>",
```
pgavlin, line 17: one question--should this be a peer of `rollback`, which is currently `esc env rollback`? or should we move `rollback` under `version`?

cleverguy25: I do think we should be consistent, I think both should be under version

cleverguy25 (approved):

cmd/esc/cli/env_version_retract.go
pgavlin: Okay. I'll follow up with a PR to move `rollback` under `version`.

PR #345, opened 2024-06-25
cli: add support for environment tags

dschaller:
add new CLI commands for CRUD operations against environment tags

```bash
> esc env tag ls localumi/env-a

This command lists all tags on the environment in the format below
Name: <tag name>
Value: <tag value>
Last updated at <timestamp> by <user>

> esc env tag localumi/env-b owner dschaller
This command adds a new tag to the specified environment and returns the new tag output similar to the format from list.
If there is already a tag with the specified key (e.g. owner) it will update the value to the new value.

Optionally, users can pass in another argument to specify an update to the key.
> esc env tag localumi/env-b owner Owner dschaller

> esc env tag rm localumi/env-b owner
This command removes a tag from the specified environment with the specified name
```


dschaller:
Also open to the idea of `esc env tag rm` and `esc env tag` update by ID both taking another argument as the tag or tag ID if it makes more sense

pgavlin:
The tag IDs are a bit surprising. We don't typically expose IDs like this--I would have expected this to act exactly like stack tags.

```console
infrastructure ❯ pulumi stack tag   
Manage stack tags

Stacks have associated metadata in the form of tags. Each tag consists of a name
and value. The `get`, `ls`, `rm`, and `set` commands can be used to manage tags.
Some tags are automatically assigned based on the environment each time a stack
is updated.

Usage:
  pulumi stack tag [command]

Available Commands:
  get         Get a single stack tag value
  ls          List all stack tags
  rm          Remove a stack tag
  set         Set a stack tag

Flags:
  -h, --help           help for tag
  -s, --stack string   The name of the stack to operate on. Defaults to the current stack

Global Flags:
      --color string                 Colorize output. Choices are: always, never, raw, auto (default "auto")
  -C, --cwd string                   Run pulumi as if it had been started in another directory
      --disable-integrity-checking   Disable integrity checking of checkpoint files
  -e, --emoji                        Enable emojis in the output (default true)
  -Q, --fully-qualify-stack-names    Show fully-qualified stack names
      --logflow                      Flow log settings to child processes (like plugins)
      --logtostderr                  Log to stderr instead of to files
      --memprofilerate int           Enable more precise (and expensive) memory allocation profiles by setting runtime.MemProfileRate
      --non-interactive              Disable interactive mode for all commands
      --profiling string             Emit CPU and memory profiles and an execution trace to '[filename].[pid].{cpu,mem,trace}', respectively
      --tracing file:                Emit tracing to the specified endpoint. Use the file: scheme to write tracing data to a local file
  -v, --verbose int                  Enable verbose logging (e.g., v=3); anything >3 is very verbose

Use "pulumi stack tag [command] --help" for more information about a command.
```

dschaller:
👍 I can change the behavior to mirror stack tags.

I intentionally made some tweaks to these commands vs the stack tags for a few reasons:
- I exposed the tag IDs to allow for updating the tag name and to handle a scenario where we potentially allow for stack tags with duplicate keys
- I kept the tag a single argument with a reserved delimiter (`:`) to allow for flexibility if we want to support bulk tagging from a single command
- Thinking more about it though I'm not really a fan of having the ID or tag name in the path so I will definitely update any of those patterns to move it out into an argument

dschaller:
@pgavlin updated these commands to more closely align with stack tags

* we no longer expose the tag ID anywhere
* the tag identifier has been moved out of the project/env path anywhere it was kept
* moved the tag key/value to be two separate arguments - I really do think we should have a way to do this as a single argument as it allows for bulk tagging but at the moment we don't have a good delimiter that cannot be used in the tag name (`:`) is a valid tag name character :(

As an aside there is also a fix to show local time if the UTC flag is not set since most of the timestamps coming back from the server are UTC

cmd/esc/cli/env_tag.go
```diff
@@ -0,0 +1,128 @@
...
+					if t.Name == name {
+						curTag = t
+						break
+					}
+				}
+
+				if after == "0" {
+					break
+				}
+			}
```
pgavlin, line 78: is there any way we can make this simpler? or is this the only option we have for getting the ID of a tag?

honestly, I'm still a little confused by the fact that tags have an ID beyond their name. does the name not need to be unique?

cmd/esc/cli/env_tag.go
```diff
@@ -0,0 +1,128 @@
...
+
+			st := style.NewStylist(style.Profile(env.esc.stdout))
+
+			if curTag != nil {
+				if curTag.Name == name && curTag.Value == value {
+					printTag(env.esc.stdout, st, curTag, utcFlag(utc))
+					return nil
+				}
```
pgavlin, line 86: how do I get the value of the tag without already knowing it? use `tag ls`?

dschaller: yeah at the moment it's only through the list command but I can add an endpoint to lookup a tag by the identifier / name which we could use in a few other places instead of fetching all of the tags

pgavlin:
Could you update the PR description with the new command structure? Also--while the example output is nice, it's a bit tough to parse. I think I'd prefer a brief plain-english description of what each command does to the output.

dschaller:
@pgavlin updated the description

cmd/esc/cli/cli_test.go
```diff
@@ -627,6 +627,86 @@ func (c *testPulumiClient) GetOpenProperty(ctx context.Context, orgName, envName
 	return nil, errors.New("NYI")
 }
 
+func (c *testPulumiClient) GetEnvironmentTag(
```
pgavlin, line 630: Can you add some tests?

If you add a new file in `cli/testdata`, it will be automatically picked up by the test harness:

- The `run` section contains the command(s) to run
- The `environments` section contains the environments to define for testing

Once you've filled in those two sections, you can generate the `stdout`/`stderr` sections by running `PULUMI_ACCEPT=1 go test ./cmd/esc/cli` from the repository root.

You'll likely need to revert changes to `cmd/esc/cli/testdata/env-login-gh100.yaml` and `cmd/esc/cli/testdata/env-set.yaml` as the regeneration process makes undesirable formatting changes to those files.

cmd/esc/cli/cli_test.go
```diff
@@ -990,8 +992,22 @@ func (c *testExec) runScript(script string, cmd *exec.Cmd) error {
...
+				esc.SetContext(ctx)
+				defer cancel()
+
+				ch := make(chan error, 1)
+				go func() {
+					ch <- esc.Execute()
+				}()
+				select {
+				case <-ctx.Done():
 					return interp.NewExitStatus(1)
```
pgavlin, line 1006: Can't add suggestions since this line didn't change, but we should return the context's error here--it'll be a bit more informative:

```golang
					return ctx.Err()
```

cmd/esc/cli/testdata/env-tag-update-key.yaml
```diff
@@ -0,0 +1,17 @@
+run: |
+  esc env tag test-org/env owner pulumi
+  esc env tag test-org/env owner team pulumi
```
pgavlin, line 3: ergonomics feel a little weird here. would prefer a `tag mv` command like

```console
> esc env tag mv test-org/env owner team
```

pgavlin (changes requested):

cmd/esc/cli/cli_test.go
```diff
@@ -1110,8 +1130,14 @@ func loadTestcase(path string) (*cliTestcaseYAML, *cliTestcase, error) {
 			revisions = cliTestcaseRevisions{Revisions: []cliTestcaseRevision{{YAML: env}}}
 		}
 
+		var tags cliTestcaseEnvironmentTags
+		envTags := map[string]string{}
+		if err := env.Decode(&tags); err == nil {
+			envTags = tags.Tags
+		}
```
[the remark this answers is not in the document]
dschaller: Sorry do you mean moving the existing tags to `cliTestcaseRevisions` instead of renaming them to `revisionTags` on the test environment?

cmd/esc/cli/testdata/env-tag-update-key.yaml
dschaller: yeah, I was trying to stick with the same pattern we employ in `esc env tag org/env@tag` since that will both create or move a tag but I can make this explicit

cmd/esc/cli/cli_test.go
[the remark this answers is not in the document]
pgavlin: Yeah--it looks like as written an environment can either contain a set of tags or a set of revisions (unless I'm reading it wrong?). Feels like instead we want `cliTestcaseRevisions` to contain a field of type `cliTestcaseEnvironmentTags`.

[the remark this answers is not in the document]
dschaller: ah yeah I follow you - will update

cmd/esc/cli/testdata/env-tag-mv.yaml
```diff
@@ -0,0 +1,24 @@
+run: |
+  esc env tag ls test-org/env
+  esc env tag mv test-org/env team esc
+  esc env tag mv test-org/env team owner pulumi
```
pgavlin, line 4: what does this do? does this rename the tag and give it a new value?

cmd/esc/cli/testdata/env-tag.yaml
```diff
@@ -0,0 +1,34 @@
+run: |
+  esc env tag ls test-org/env
+  esc env tag test-org/env owner pulumi
+  esc env tag get test-org/env owner
```
pgavlin, line 4: do we need a separate `get` command? could `tag` w/o a value just return the current value of the tag?

cmd/esc/cli/testdata/env-tag-mv.yaml
dschaller: 👍

cmd/esc/cli/testdata/env-tag.yaml
dschaller: It's not required here since the return value of `tag` does that but I added it to the test to demonstrate that it is being persisted

cmd/esc/cli/cli_test.go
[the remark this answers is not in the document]
dschaller: actually thinking more about this I feel like it makes sense to keep these at the top level on the env since revision tags must be unique across all revisions. I did adjust the interface of revision tags though to allow for passing a list instead of a single value since that mirrors what the API supports

cmd/esc/cli/env_tag_mv.go
```diff
@@ -0,0 +1,80 @@
...
+
+	"github.com/pulumi/esc/cmd/esc/cli/style"
+	"github.com/spf13/cobra"
+)
+
+func newEnvTagMvCmd(env *envCommand) *cobra.Command {
+	var utc bool
+
+	cmd := &cobra.Command{
+		Use:   "mv [<org-name>/]<environment-name> <name> [<newName>] <value>",
```
pgavlin, line 17: I would remove the `<value>` bit here and make this command purely a "rename" command:

`mv [<org-name>/]<environment-name> <name> <newName>`

cmd/esc/cli/env_tag.go
```diff
@@ -0,0 +1,90 @@
...
+	"github.com/pulumi/esc/cmd/esc/cli/style"
+	"github.com/spf13/cobra"
+)
+
+func newEnvTagCmd(env *envCommand) *cobra.Command {
+	var utc bool
+
+	cmd := &cobra.Command{
+		Use:   "tag [<org-name>/]<environment-name> <name> <value>",
+		Args:  cobra.ExactArgs(3),
```
pgavlin, line 21: should we push this functionality under a `set` command?

cmd/esc/cli/env_tag_mv.go
dschaller: ah ok and keep the value updates under the tag "set" command?

dschaller: added back the update logic to the implied "set" command

cmd/esc/cli/env_tag_mv.go
```diff
@@ -0,0 +1,74 @@
...
+
+	"github.com/pulumi/esc/cmd/esc/cli/style"
+	"github.com/spf13/cobra"
+)
+
+func newEnvTagMvCmd(env *envCommand) *cobra.Command {
+	var utc bool
+
+	cmd := &cobra.Command{
+		Use:   "mv [<org-name>/]<environment-name> <name> <newName>",
```
pgavlin, line 17:
```suggestion
		Use:   "mv [<org-name>/]<environment-name> <name> <new-name>",
```

cmd/esc/cli/cli_test.go
```diff
@@ -253,13 +253,14 @@ type testEnvironmentRetract struct {
 type testEnvironmentRevision struct {
 	number    int
 	yaml      []byte
-	tag       string
+	tags      map[string]bool
```
pgavlin, line 256: curious why this changed? this is tracking the revision's etag, which should be static?

dschaller: Oh interesting - this was used by the test files as a way to pass in a revision tag so I modified it to take multiple tags to mirror production

pgavlin: > I did notice it was also being used as the etag but I figured that was solely out of convenience

ah interesting--this double duty was unintentional, and this should be the etag. for convenience you can leave these changes if you like and I can straighten things out later

pgavlin (approved):

PR #353, opened 2024-07-16
[cli] Stringify scalar secrets

### LINKED ISSUE #296: Cannot Set Numerical Only Secrets
### What happened?

I'm trying to save secrets that only have numerical values

### Example

```sh
$ esc env set dosu/staging --secret pulumiConfig.app:GH_APP_ID '1234'
Error: updating environment definition: [0]
Diags: secret values must be string literals
```

### Output of `pulumi about`

CLI
Version      3.114.0
Go Version   go1.22.2
Go Compiler  gc

Plugins
KIND      NAME  VERSION
resource  gcp   7.19.0
language  go    unknown

Host
OS       darwin
Version  14.4.1
Arch     arm64

This project is written in go: executable='/opt/homebrew/bin/go' version='go version go1.21.3 darwin/arm64'

Backend
Name           pulumi.com
URL            https://app.pulumi.com/devstein
User           devstein
Organizations  devstein, src, dosu
Token type     personal

Dependencies:
NAME                                 VERSION
github.com/pulumi/pulumi-gcp/sdk/v7  v7.19.0
github.com/pulumi/pulumi/sdk/v3      v3.113.3

Pulumi locates its logs in /var/folders/wq/7c2lyfl92v92vy__dqyhbkbc0000gn/T/ by default

### Additional context

I'm able to save numerical config values, but not secrets

If you'd like help triaging issues like these try out [Dosu](https://github.com/apps/dosubot). I built it because I've spent too much time on GitHub issues 😁 

### Contributing

pgavlin:
`esc env set --secret` requires that its argument is a string literal. While this is consistent with the requirements of the evaluator, it is somewhat less than ergonomic for CLI users, who must surround numeric or boolean values in both single and double quotes:

```console
esc env set my-env --secret foo.bar '"1234"'
```

These changes detect non-string scalar values passed to `esc env set --secret` and automatically coerce these values to strings. Non-scalar values are passed through to the evaluator as-is, which will report that the argument must be a string literal.

Fixes #296.

### DIFF
--- a/cmd/esc/cli/env_set.go
+++ b/cmd/esc/cli/env_set.go
@@ -6,6 +6,7 @@ import (
 	"context"
 	"fmt"
 	"regexp"
+	"strconv"
 
 	"github.com/ccojocar/zxcvbn-go"
 	"github.com/spf13/cobra"
@@ -71,6 +72,14 @@ func newEnvSetCmd(env *envCommand) *cobra.Command {
 				return fmt.Errorf("value looks like a secret; rerun with --secret to mark it as such, or --plaintext if you meant to leave it as plaintext")
 			}
 			if secret {
+				if yamlValue.Kind == yaml.ScalarNode && yamlValue.Tag != "!!str" {
+					err = yaml.Unmarshal([]byte(strconv.Quote(args[1])), &yamlValue)
+					if err != nil {
+						return fmt.Errorf("internal error decoding value; try surrounding the argument in both single and double quotes (e.g. '\"foo\"') (%w)", err)
+					}
+					yamlValue = *yamlValue.Content[0]
+				}
+
 				bytes, err := yaml.Marshal(map[string]yaml.Node{"fn::secret": yamlValue})
 				if err != nil {
 					return fmt.Errorf("internal error: marshaling secret: %w", err)
--- a/cmd/esc/cli/testdata/env-set-secret.yaml
+++ b/cmd/esc/cli/testdata/env-set-secret.yaml
@@ -8,7 +8,16 @@ run: |
   esc env get test password --show-secrets
   esc env set test password YMHdog6VkUqbVS5dAErxWeMy83r4HeBeNRdEuUFwK3M= --plaintext
   esc env get test
-stdout: |+
+  esc env set test password 1234 --secret
+  esc env get test
+  esc env set test password 123.4 --secret
+  esc env get test
+  esc env set test password false --secret
+  esc env get test
+  esc env set test password true --secret
+  esc env get test
+  esc env set test password '[]' --secret || echo OK
+stdout: |
   > esc env init test
   Environment created.
   > esc env set test password YMHdog6VkUqbVS5dAErxWeMy83r4HeBeNRdEuUFwK3M=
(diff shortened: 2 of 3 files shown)


cmd/esc/cli/env_set.go
```diff
@@ -71,6 +73,14 @@ func newEnvSetCmd(env *envCommand) *cobra.Command {
 				return fmt.Errorf("value looks like a secret; rerun with --secret to mark it as such, or --plaintext if you meant to leave it as plaintext")
 			}
 			if secret {
+				if yamlValue.Kind == yaml.ScalarNode && yamlValue.Tag != "!!str" {
+					err = yaml.Unmarshal([]byte(strconv.Quote(args[1])), &yamlValue)
+					if err != nil {
+						return errors.New("secret values must be strings; try surrounding the argument in both single and double quotes (e.g. '\"foo\"')")
```
lunaris, line 79: Will this change anything? Given that presumably, you have just done this internally (quoted the string). Appreciate the point of the error might be just to get someone to re-evaluate the argument they have passed, which if so fair enough. That said, if we're doing that -- perhaps include the value they passed in the error message?

lunaris (approved):
Not sure if I have the powah :tm: here, but this looks sensible to me.

cmd/esc/cli/env_set.go
lblackstone, line 79: This error message is confusing to me since the point of the PR is to coerce non-string values into strings. It looks like the test case for this is `[]`, which you could quote, but seems like a type error. I'm not sure what to suggest instead.

lblackstone (approved):
Code change LGTM aside from comment on the error message.

cmd/esc/cli/env_set.go
pgavlin: See my response to @lblackstone below--this is a bit of paranoia on my part to handle the case where auto-quoting doesn't work properly.

pgavlin: So this error message is a bit of paranoia on my part.

We only hit this branch if the value the user provided deserializes as a scalar, so it doesn't trip on arrays or objects. So we'd never reach here on `[]`. This guards against the transformation being performed by `strconv.Quote` going wrong in some way, and points the user towards a potential solution.

We could instead issue an internal error and direct the user to file a bug.

pgavlin: thoughts @lunaris @lblackstone ?

lblackstone: > We could instead issue an internal error and direct the user to file a bug.

I like this suggestion better. Ideally, we could print the error message from `yaml.Unmarshal`, but would want to avoid printing out the plaintext secret, so something more generic would be fine as well.

pgavlin: Done

cmd/esc/cli/testdata/env-set-secret.yaml
```diff
@@ -100,3 +178,13 @@ stderr: |
...
+  > esc env set test password 1234 --secret
+  > esc env get test
+  > esc env set test password 123.4 --secret
+  > esc env get test
+  > esc env set test password false --secret
+  > esc env get test
+  > esc env set test password true --secret
+  > esc env get test
+  > esc env set test password [] --secret
+  test:3:21: secret values must be string literals
```
lblackstone, line 190: Is this error message still accurate?

pgavlin: Roughly, yes. If you're editing the YAML document directly, secret values must be string literals. On the CLI we allow you to fudge and present non-string scalars.

Issue #347, opened 2024-07-01
Getting a secret from an imported environment returns `[unknown]`


pierskarsenbarg:
### What happened?

When trying to retrieve a config value from one environment that was imported through another, the value returned was `[unknown]`

### Example

(Commands below are using the Pulumi CLI, but the one that returns the `[unknown]` value is the same with the ESC CLI.)

Follow steps:

1. Set up `demo-base` environment: `pulumi env init demo-base`
2. Set up `demo-dev` environment: `pulumi env init demo-dev`
3. Add the following to the `demo-dev` environment:

```
imports:
    - demo-base
```

4. Add secret to demo-base environment: `pulumi env set demo-base pulumiConfig.myBasePassword "my-base-password" --secret`
5. Add secret to demo-dev environment: `pulumi env set demo-dev pulumiConfig.myDevPassword "my-dev-password" --secret`
6. Get base secret from base environment: `pulumi env get demo-base pulumiConfig.myBasePassword --show-secrets --value json`
7. Get dev secret from dev environment: `pulumi env get demo-dev pulumiConfig.myDevPassword --show-secrets --value json`
8. Try to get base secret from dev environment: `pulumi env get demo-dev pulumiConfig.myBasePassword --show-secrets --value json`

(Will return `[unknown]`)

if you run `pulumi env open demo-dev` that works correctly and the following code works as expected:

```typescript
import * as pulumi from "@pulumi/pulumi";

const config = new pulumi.Config();

export const basePw = config.get("myBasePassword");
export const devPw = config.get("myDevPassword");
```

when `Pulumi.dev.yaml` is as follows:

```yaml
environment:
  - demo-dev
```

This is what my `demo-base` environment looks like:

```yaml
values:
  pulumiConfig:
    myBasePassword:
      fn::secret:
        ciphertext: ZXNjeAAAAAEAAAEA28KPa3Jj1chaQtbF+Hahccp22+0BMvymrx0ta5PzEX3/yKVZCwJ044qt4CSHhK6h

```

this is what my `demo-dev` environment looks like:

```yaml
imports:
  - demo-base
values:
  pulumiConfig:
    myDevPassword:
      fn::secret:
        ciphertext: ZXNjeAAAAAEAAAEAHuH2JF7Gz4ad/e3ZTfR63nN6AA2G+Llrktj2HDXAvfia57zbhR0XTaQqd+oetes=

```

[...]

PR #355, opened 2024-07-17
cli: deeper support for show-secrets in env get 

pgavlin:
These changes plumb support for showing _all_ static secrets through
`esc env get` (rather than only those secrets immediately present in
the environment being read directly; i.e. including secrets from
imported environments).

Note that this support is latent, and the fix will not be user-visible
until supporting code is deployed to the ESC API.

Part of https://github.com/pulumi/esc/issues/347.

### DIFF
--- a/cmd/esc/cli/client/client.go
+++ b/cmd/esc/cli/client/client.go
@@ -42,7 +42,18 @@ import (
 const etagHeader = "ETag"
 const revisionHeader = "Pulumi-ESC-Revision"
 
+type CheckYAMLOption struct {
+	ShowSecrets bool
+}
+
 // Client provides a slim wrapper around the Pulumi HTTP/REST API.
+//
+// NOTE: this is not considered a public API, and we reserve the right to make breaking changes, including adding
+// parameters, removing methods, changing types, etc.
+//
+// However, there is currently a cyclic dependency between the Pulumi CLI and the ESC CLI that causes breaking changes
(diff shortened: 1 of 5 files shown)


pgavlin:
Depends on https://github.com/pulumi/esc/pull/353. Also--the concrete changes to the shipping CLI are latent, and will not take effect until the ESC API is updated to respect the new query parameter.

cmd/esc/cli/client/client.go
```diff
@@ -133,6 +137,7 @@ type Client interface {
 		ctx context.Context,
 		orgName string,
 		yaml []byte,
+		opts ...CheckYAMLOption,
```
lblackstone, line 140: Any particular reason to make this a variadic arg rather than a single option struct?

pgavlin: *sigh*

Yes. So because of reasons, we can't add required parameters to any methods that are called by the Pulumi CLI. Adding required parameters breaks the ESC CLI build due to a cyclic dependency on the Pulumi CLI.

Ideally this would just be a `bool`.

lblackstone: Ouch. Probably worth adding a comment to that effect.

lblackstone (approved):

cleverguy25 (approved):

cmd/esc/cli/client/client.go
pgavlin: Done. Added a comment to the interface as well.

Issue #298, opened 2024-05-08
Panic when editing environment


komalali:
### What happened?

Panic:
```
/usr/local/go/src/reflect/value.go:247 +0xb8
reflect.flag.mustBeExported(...)
    /usr/local/go/src/reflect/value.go:241
reflect.Value.Set({0x2bf3a40?, 0x4004e82b08?, 0x2bf3a40?}, {0x0?, 0x0?, 0x400427cfc0?})
    /usr/local/go/src/reflect/value.go:2320 +0x88
[github.com/pulumi/esc/ast.parseField(](http://github.com/pulumi/esc/ast.parseField(){0x4002c21b80, 0x5}, {0x2bf3a40?, 0x4004e82b08?, 0x4002af71c8?}, {0x3bd0900, 0x400427c9a0})
    /go/src/github.com/pulumi/pulumi-service/vendor/github.com/pulumi/esc/ast/environment.go:303 +0x930
[github.com/pulumi/esc/ast.parseRecord({0x31c12ed](http://github.com/pulumi/esc/ast.parseRecord(%7B0x31c12ed), 0x6}, {0x3bad760, 0x4004e82af8}, {0x3bd08c8?, 0x400313b050?}, 0x0)
    /go/src/github.com/pulumi/pulumi-service/vendor/github.com/pulumi/esc/ast/environment.go:326 +0x38c
[github.com/pulumi/esc/ast.(*ImportDecl).parse(0x400427cf80](http://github.com/pulumi/esc/ast.(*ImportDecl).parse(0x400427cf80), {0x400427cf80?, 0x2b420a0?}, {0x3bd08c8?, 0x400313b170?})
    /go/src/github.com/pulumi/pulumi-service/vendor/github.com/pulumi/esc/ast/environment.go:178 +0x2b8
[github.com/pulumi/esc/ast.parseField({0x4002c21af0](http://github.com/pulumi/esc/ast.parseField(%7B0x4002c21af0), 0xa}, {0x2bf3cc0?, 0x4003142d88?, 0x2?}, {0x3bd08c8, 0x400313b170…
Message
[request id: 4bb7447d-014c-43bc-a863-ceb86f3f39be reflect: call of reflect.Value.Set on zero Value]
```

### Example

The repro is to try adding `merge: false` to an import.
Specifically, I was in the providers.all.unmerged environment and after aws I hit return, tab, started typing merge: false and the 500 happens after the colon.

### Output of `pulumi about`

n/a

### Contributing

[...]

PR #359, opened 2024-07-18
Initialize field before use

lblackstone:
The `parseField` function was not initializing `v` on an error case, which could led to subsequent code trying to use an uninitialized pointer and causing a panic.

Fix https://github.com/pulumi/esc/issues/298

### DIFF
--- a/ast/environment.go
+++ b/ast/environment.go
@@ -293,7 +293,7 @@ func parseField(name string, dest reflect.Value, node syntax.Node) syntax.Diagno
 		xv := reflect.ValueOf(x)
 		if !xv.Type().AssignableTo(dest.Type()) {
 			diags.Extend(exprFieldTypeMismatchError(name, dest.Interface(), x))
-			return diags
+			v = reflect.New(dest.Type().Elem())
 		} else {
 			v = xv
 		}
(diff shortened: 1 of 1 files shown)


pgavlin:
Can you post an example of the diag message that we now present in the case of the linked issue?

ast/environment.go
```diff
@@ -293,6 +293,7 @@ func parseField(name string, dest reflect.Value, node syntax.Node) syntax.Diagno
 		xv := reflect.ValueOf(x)
 		if !xv.Type().AssignableTo(dest.Type()) {
 			diags.Extend(exprFieldTypeMismatchError(name, dest.Interface(), x))
+			return diags
```
pgavlin, line 296: The only concern I have about an early return here is that it leaves `dest` uninitialized. Instead, we might want to create a new zero value of the destination type and fall through. This would be consistent with the other branches.

I think that's just `v = reflect.New(dest.Type().Elem())`.

lblackstone:
```
<yaml>:4:13 - imports[0].test.merge:
		Error: merge must be a boolean value
```

Before
https://github.com/user-attachments/assets/ff1b2335-1dae-448c-9e4e-c850b53f092c

After
https://github.com/user-attachments/assets/34fa5762-6071-4acb-8d21-37446d0f3518

ast/environment.go
lblackstone: Updated

pgavlin (approved):

PR #364, opened 2024-07-25
eval: fix showSecrets in CheckEnvironment

pgavlin:
These changes fix a bug in `CheckEnvironment` when `showSecrets` is
`false`. In this case, if the environment loader supplied a decrypter
for an imported environment, the imported environment's secrets would be
decrypted, violating the `showSecrets` contract. In order to address
this issue, these changes store and propagate the value of `showSecrets`
in the evaluator's context.

### DIFF
--- a/eval/eval.go
+++ b/eval/eval.go
@@ -147,7 +147,7 @@ type imported struct {
 type evalContext struct {
 	ctx          context.Context      // the cancellation context for evaluation
 	validating   bool                 // true if we are only checking the environment
-	showSecrets  bool                 // true if serets should be decrypted during validation
+	showSecrets  bool                 // true if secrets should be decrypted during validation
 	name         string               // the name of the environment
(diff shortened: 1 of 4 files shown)


pgavlin:
These changes are structured as two commits: one that adds a test case to demonstrate the existing issue and one that fixes the issue and updates the test case.

eval/eval.go
```diff
@@ -149,6 +147,7 @@ type imported struct {
 type evalContext struct {
 	ctx          context.Context      // the cancellation context for evaluation
 	validating   bool                 // true if we are only checking the environment
+	showSecrets  bool                 // true if serets should be decrypted during validation
```
lblackstone, line 150:
```suggestion
	showSecrets  bool                 // true if secrets should be decrypted during validation
```

lblackstone (approved):

PR #368, opened 2024-08-02
Handle login.Current returning nil

Frassle:
Fixes https://github.com/pulumi/esc/issues/367.

If the pulumi credentials have a current account saved, but that account's access key has expired `httpstate.Current` will return `nil, nil`. `getCachedCredentials` did not correctly handle this `nil, nil` result.

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -8,3 +8,5 @@
 
 ### Bug Fixes
 
+- Fix a panic in fetching current credentials when the access key had expired.
+  [#368](https://github.com/pulumi/esc/pull/368)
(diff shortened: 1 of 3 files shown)


pgavlin:
What are the circumstances under which this happens? Could you add a note to that effect to the PR description?

It'd also be nice to have a quick unit test. The various bits in here are mockable (e.g. the login manager), so it should be pretty straightforward to roll one together.

Frassle:
> What are the circumstances under which this happens?

I think this is from an account being saved in the pulumi credentials but it's access_key having expired.

pgavlin (approved):

PR #369, opened 2024-08-07
Add Projects

seanyeh:
## Description
Adds Projects (groupings of ESC environments) to the ESC CLI

## Testing
setup: personal/default org = `syeh2`. other orgs = `ambiguous`

### Init
```bash
❯ go run ./cmd/esc env init env1
Environment created: syeh2/default/env1

❯ go run ./cmd/esc env init new-project/env1
Environment created: syeh2/new-project/env1

# Legacy pattern (org/env) because I have an org called `ambiguous`
❯ go run ./cmd/esc env init ambiguous/myenv
Environment created: ambiguous/default/myenv

# Test creating a project with the same name as an org
❯ go run ./cmd/esc env init syeh2/ambiguous/env1
Environment created: syeh2/ambiguous/env1

# Because now I have an existing env under the `ambiguous` project, I can create a new env under the same project with just `ambiguous/env2`
❯ go run ./cmd/esc env init ambiguous/env2
Environment created: syeh2/ambiguous/env2

```

### Get
Setup: Created 'syeh2/ambiguous/myenv' and 'ambiguous/default/myenv` envs beforehand
```bash
❯ go run ./cmd/esc env get ambiguous/myenv --value json
Error: ambiguous path provided

Environments found at both 'syeh2/ambiguous/myenv' and 'ambiguous/default/myenv'.
Please specify the full path as <org-name>/<project-name>/<env-name>
exit status 1

❯ go run ./cmd/esc env get syeh2/ambiguous/myenv --value json
{
...
}

❯ go run ./cmd/esc env get ambiguous/default/myenv --value json
{
...
}
```

### Get with cross-project imports
Setup: created proj1/shared, proj2/env
```bash
❯ go run ./cmd/esc env get proj2/env

   Value

    {
      "a": "hello"
    }

   Definition

    imports:
      - proj1/shared


```

### Edit/open
```bash
❯ go run ./cmd/esc env edit new-project/env1
Environment updated.

❯ go run ./cmd/esc env open new-project/env1
{
  "a": "hello"
}
```
### Rm
```bash
❯ go run ./cmd/esc env rm new-project/env1
This will permanently remove the "syeh2/new-project/env1" environment!
Please confirm that this is what you'd like to do by typing `syeh2/new-project/env1`: syeh2/new-project/env1
Environment "syeh2/new-project/env1" has been removed!
```

### Diff
Setup: created 'test2/default/versioned' with the `old` tag pointing to an older version
```bash
❯ go run ./cmd/esc env diff test2/versioned@old -f json
--- test2/default/versioned@old
+++ test2/default/versioned@latest
@@ -1,3 +1,3 @@
 {
-  "example": "first"
+  "example": "newer"
 }

❯ go run ./cmd/esc env diff test2/versioned@old test2/versioned@old -f json
# empty because diffing against itself

❯ go run ./cmd/esc env diff test2/versioned@old test2/versioned@latest -f json
--- test2/default/versioned@old
+++ test2/default/versioned@latest
@@ -1,3 +1,3 @@
 {
-  "example": "first"
+  "example": "newer"
 }

❯ go run ./cmd/esc env diff test2/versioned@old test2/default/versioned@latest -f json
--- test2/default/versioned@old
+++ test2/default/versioned@latest
@@ -1,3 +1,3 @@
 {
-  "example": "first"
+  "example": "newer"
 }
```

### Version
```bash
❯ go run ./cmd/esc env version test2/test-project/envvars@latest
latest
Revision 2
Last updated at 2024-08-09 08:45:52.766 -0500 CDT by  <syeh2>

❯ go run ./cmd/esc env version history test2/test-project/envvars
revision 2 (tag: latest)
Author:  <syeh2>
Date:   2024-08-09 08:45:52.765 -0500 CDT

revision 1
Author:  <syeh2>
Date:   2024-08-09 08:44:34.019 -0500 CDT
```

### DIFF
--- a/cmd/esc/cli/cli_test.go
+++ b/cmd/esc/cli/cli_test.go
@@ -788,13 +788,14 @@ func (c *testPulumiClient) DeleteEnvironmentTag(ctx context.Context, orgName, pr
 func (c *testPulumiClient) GetEnvironmentRevision(
 	ctx context.Context,
 	orgName string,
+	projectName string,
 	envName string,
 	revision int,
 ) (*client.EnvironmentRevision, error) {
 	before, count := revision+1, 1
 
 	opts := client.ListEnvironmentRevisionsOptions{Before: &before, Count: &count}
-	revs, err := c.ListEnvironmentRevisions(ctx, orgName, envName, opts)
+	revs, err := c.ListEnvironmentRevisions(ctx, orgName, projectName, envName, opts)
 	if err != nil || len(revs) == 0 {
 		return nil, err
 	}
@@ -804,6 +805,7 @@ func (c *testPulumiClient) GetEnvironmentRevision(
 func (c *testPulumiClient) ListEnvironmentRevisions(
 	ctx context.Context,
 	orgName string,
+	projectName string,
 	envName string,
 	options client.ListEnvironmentRevisionsOptions,
 ) ([]client.EnvironmentRevision, error) {
@@ -854,6 +856,7 @@ func (c *testPulumiClient) ListEnvironmentRevisions(
 func (c *testPulumiClient) RetractEnvironmentRevision(
 	ctx context.Context,
 	orgName string,
+	projectName string,
 	envName string,
 	version string,
 	replacement *int,
--- a/cmd/esc/cli/cli_test.go
+++ b/cmd/esc/cli/cli_test.go
@@ -236,8 +236,17 @@ type testEnvironments struct {
 	environments map[string]*testEnvironment
 }
 
-func (e *testEnvironments) LoadEnvironment(ctx context.Context, envName string) ([]byte, eval.Decrypter, error) {
-	name := path.Join(e.orgName, envName)
+func (e *testEnvironments) LoadEnvironment(ctx context.Context, ref string) ([]byte, eval.Decrypter, error) {
+	var name string
+
+	// This "emulates" the backend behavior of resolving refs
+	if strings.Contains(ref, "/") {
+		name = path.Join(e.orgName, ref)
+	} else {
+		// If ref is a single identifier, assume the default project
+		name = path.Join(e.orgName, client.DefaultProject, ref)
+	}
+
 	env, ok := e.environments[name]
 	if !ok {
 		return nil, nil, errors.New("not found")
@@ -997,6 +1006,16 @@ func (c *testPulumiClient) ListEnvironmentRevisionTags(
 	return fx.ToSlice(fx.FMap(fx.IterSlice(names), func(name string) (client.EnvironmentRevisionTag, bool) {
 		return client.EnvironmentRevisionTag{Name: name, Revision: env.revisionTags[name]}, name > options.After
 	})), nil
+}
+
+// unused
+func (c *testPulumiClient) EnvironmentExists(
+	ctx context.Context,
+	orgName string,
+	projectName string,
+	envName string,
+) (bool, error) {
+	return false, nil
 }
 
 type testExec struct {
--- a/cmd/esc/cli/cli_test.go
+++ b/cmd/esc/cli/cli_test.go
@@ -348,8 +348,8 @@ func mapDiags(diags syntax.Diagnostics) []client.EnvironmentDiagnostic {
 	return out
 }
 
-func (c *testPulumiClient) getEnvironment(orgName, envName, version string) (*testEnvironment, *testEnvironmentRevision, error) {
-	name := path.Join(orgName, envName)
+func (c *testPulumiClient) getEnvironment(orgName, projectName, envName, version string) (*testEnvironment, *testEnvironmentRevision, error) {
+	name := path.Join(orgName, projectName, envName)
 
 	env, ok := c.environments[name]
 	if !ok {
@@ -451,8 +451,8 @@ func (c *testPulumiClient) GetPulumiAccountDetails(ctx context.Context) (string,
 	return c.user, nil, nil, nil
 }
 
-func (c *testPulumiClient) GetRevisionNumber(ctx context.Context, orgName, envName, version string) (int, error) {
-	_, rev, err := c.getEnvironment(orgName, envName, version)
+func (c *testPulumiClient) GetRevisionNumber(ctx context.Context, orgName, projectName, envName, version string) (int, error) {
+	_, rev, err := c.getEnvironment(orgName, projectName, envName, version)
 	if err != nil {
 		return 0, err
 	}
@@ -520,7 +520,7 @@ func (c *testPulumiClient) GetEnvironment(
 	version string,
 	showSecrets bool,
 ) ([]byte, string, int, error) {
-	_, env, err := c.getEnvironment(orgName, envName, version)
+	_, env, err := c.getEnvironment(orgName, projectName, envName, version)
 	if err != nil {
 		return nil, "", 0, err
 	}
@@ -560,18 +560,19 @@ func (c *testPulumiClient) UpdateEnvironmentWithProject(
 	yaml []byte,
 	tag string,
 ) ([]client.EnvironmentDiagnostic, error) {
-	diags, _, err := c.UpdateEnvironmentWithRevision(ctx, orgName, envName, yaml, tag)
+	diags, _, err := c.UpdateEnvironmentWithRevision(ctx, orgName, projectName, envName, yaml, tag)
 	return diags, err
 }
 
 func (c *testPulumiClient) UpdateEnvironmentWithRevision(
 	ctx context.Context,
 	orgName string,
+	projectName string,
 	envName string,
 	yaml []byte,
 	tag string,
 ) ([]client.EnvironmentDiagnostic, int, error) {
-	env, latest, err := c.getEnvironment(orgName, envName, "")
+	env, latest, err := c.getEnvironment(orgName, projectName, envName, "")
 	if err != nil {
 		return nil, 0, err
 	}
@@ -620,7 +621,7 @@ func (c *testPulumiClient) OpenEnvironment(
 	version string,
 	duration time.Duration,
 ) (string, []client.EnvironmentDiagnostic, error) {
-	_, env, err := c.getEnvironment(orgName, envName, version)
+	_, env, err := c.getEnvironment(orgName, projectName, envName, version)
 	if err != nil {
 		return "", nil, err
 	}
@@ -658,7 +659,7 @@ func (c *testPulumiClient) GetOpenEnvironmentWithProject(ctx context.Context, or
 	return env, nil
 }
 
-func (c *testPulumiClient) GetOpenProperty(ctx context.Context, orgName, envName, openEnvID, property string) (*esc.Value, error) {
+func (c *testPulumiClient) GetOpenProperty(ctx context.Context, orgName, projectName, envName, openEnvID, property string) (*esc.Value, error) {
 	return nil, errors.New("NYI")
 }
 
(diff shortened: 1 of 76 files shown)


cmd/esc/cli/cli_test.go
```diff
@@ -968,6 +1000,16 @@ func (c *testPulumiClient) ListEnvironmentRevisionTags(
 	})), nil
 }
 
+// unused
```
dschaller, line 1003: Why is this unused?

cmd/esc/cli/client/client.go
```diff
@@ -78,8 +79,10 @@ type Client interface {
 		continuationToken string,
 	) (environments []OrgEnvironment, nextToken string, err error)
 
-	// CreateEnvironment creats an environment named envName in orgName.
+	// Deprecated
```
dschaller, line 82: We should probably still keep the original docstring while this is being deprecated.

Also, nit: we should probably follow golangs deprecation format of `Deprecated: Use {x} instead.`

cmd/esc/cli/client/client.go
```diff
@@ -105,23 +109,33 @@ type Client interface {
 	UpdateEnvironmentWithRevision(
 		ctx context.Context,
 		orgName string,
+		projectName string,
 		envName string,
 		yaml []byte,
 		etag string,
 	) ([]EnvironmentDiagnostic, int, error)
 
+	// Deprecated
```
dschaller, line 118: Same nit here on deprecation format

cmd/esc/cli/client/client.go
```diff
@@ -407,26 +435,32 @@ func (pc *client) ListEnvironments(
...
 }
 
+// Deprecated
 func (pc *client) CreateEnvironment(ctx context.Context, orgName, envName string) error {
-	path := fmt.Sprintf("/api/preview/environments/%v/%v", orgName, envName)
+	return pc.CreateEnvironmentWithProject(ctx, orgName, DefaultProject, envName)
+}
+
+func (pc *client) CreateEnvironmentWithProject(ctx context.Context, orgName, projectName, envName string) error {
+	path := fmt.Sprintf("/api/esc/environments/%v/%v/%v", orgName, projectName, envName)
```
dschaller, line 451: Do we plan on renaming this after the pulumi CLI update, once the deprecated function can be removed or are we always going to stick with `CreateEnvironmentWithProject`? If the latter, maybe we can call this something like `CreateProjectEnvironment` or `CreateEnvironmentInProject` since the environment will belong to the project

cmd/esc/cli/env.go
```diff
@@ -65,49 +66,192 @@ func newEnvCmd(esc *escCommand) *cobra.Command {
...
 	}
 
-	return environmentRef{orgName, envName, version}
+	return envRef, nil
 }
 
-func (cmd *envCommand) getEnvRef(args []string) (ref environmentRef, rest []string, err error) {
+// Get an environment reference when creating a new environment
+// If the given path is ambiguous, we need to make an additional API call to disambiguate
+func (cmd *envCommand) getNewEnvRef(
```
dschaller, line 145: can we make it more explicit what the difference is between these two functions?

cmd/esc/cli/env.go
```diff
@@ -65,49 +66,192 @@ func newEnvCmd(esc *escCommand) *cobra.Command {
...
+		orgName:          orgName,
+		projectName:      projectName,
+		envName:          envName,
+		version:          version,
+		hasAmbiguousPath: hasAmbiguousPath,
+	}
+}
+
+// If a non-nil environmentRef is provided, default to its values if an environment name and version are not specified
+func (cmd *envCommand) getEnvRef(refString string, rel *environmentRef) (ref environmentRef, err error) {
```
dschaller, line 131: Is there a reason to have an error in the return values since it doesn't look like we will ever return one

cmd/esc/cli/env.go
```diff
@@ -65,49 +66,192 @@ func newEnvCmd(esc *escCommand) *cobra.Command {
...
 			return environmentRef{}, nil, fmt.Errorf("no environment name specified")
 		}
 		cmd.envNameFlag, args = args[0], args[1:]
 	}
 
-	return cmd.getRelativeEnvRef(cmd.envNameFlag, nil), args, nil
+	ref, err = cmd.getEnvRef(cmd.envNameFlag, nil)
+	if err != nil {
+		return ref, args, err
+	}
```
dschaller, line 159: We will never hit this conditional

cmd/esc/cli/env.go
```diff
@@ -65,49 +66,192 @@ func newEnvCmd(esc *escCommand) *cobra.Command {
...
+
+	useLegacyPath := false
+	_, orgs, _, _ := cmd.esc.client.GetPulumiAccountDetails(ctx)
+	for _, org := range orgs {
+		if strings.EqualFold(legacyRef.orgName, org) {
+			useLegacyPath = true
+			break
+		}
+	}
```
dschaller, line 181: what if a user has an project that is the same name as their org?

cmd/esc/cli/env.go
seanyeh: nice catch, I'll fix that. this was some vestigial artifact from refactoring

cmd/esc/cli/client/client.go
seanyeh: I would prefer removing the deprecated function and renaming, if we're okay with breaking the esc package api

cmd/esc/cli/env.go
seanyeh: We don't currently have an API for listing all projects for a user, so I went with the simple decision of just going with `<org>/<env>` if 2 identifiers are provided and the first matches a user's org. Is this reasonable? If not, we can either build out the API to list all user's projects or do something like get all of the user's environments and getting the projects from that.

cmd/esc/cli/cli_test.go
seanyeh: The test client doesn't use it. I'm still adding some tests so this may or may not change

cmd/esc/cli/env.go
```diff
@@ -65,49 +66,186 @@ func newEnvCmd(esc *escCommand) *cobra.Command {
...
+	return environmentRef{
+		orgName:          orgName,
+		projectName:      projectName,
+		envName:          envName,
+		version:          version,
+		hasAmbiguousPath: hasAmbiguousPath,
 	}
+}
+
+// If a non-nil environmentRef is provided, default to its values if an environment name and version are not specified
```
dschaller, line 130:
```suggestion
// If a non-nil environmentRef is provided, default to its values if an environment name is not specified but a version is
```

cmd/esc/cli/env.go
```diff
@@ -65,49 +66,186 @@ func newEnvCmd(esc *escCommand) *cobra.Command {
...
+		}
+		cmd.envNameFlag, args = args[0], args[1:]
+	}
+
+	envRef, err := cmd.getExistingEnvRefWithRelative(ctx, cmd.envNameFlag, nil)
+
+	return envRef, args, err
+}
+
+func (cmd *envCommand) getExistingEnvRefWithRelative(
```
dschaller, line 205: The naming here is a bit confusing imo I think the original intention of the ref passed in was to be a way to resolve a env reference string without an env name but a version specified. WithRelative is a bit unclear how the relative is being used. Even the existing function name is a bit unclear to me since the reference returned is not always relative to the reference passed in (e.g. if the ref string contains an environment name)

cmd/esc/cli/env.go
seanyeh: nice catch. I actually had misread the condition here

cmd/esc/cli/env.go
seanyeh: Yah I can see it's confusing. `getExistingEnvRef` refers to a reference for an existing env, as opposed to `getNewEnvRef` which refers to a reference for a new env that is about to be created. (We differentiate between these 2 cases because of differing logic for checking for the legacy pattern).

Would `getRelativeExistingEnvRef` be better? or `getExistingEnvRefWithFallback`?

cmd/esc/cli/client/client.go
seanyeh: yup that sounds right

cmd/esc/cli/env.go
dschaller: I think we will want to delineate between these two because it could be unintuitive that if you specify `{project}/{env}` expecting the org to be implicit similar to other commands they would be confused that the env would be created in the default project. Also, we don't want to encourage users to create new users in the default project since it will be read only.

Having an API to list the projects probably would make sense in that it can be used elsewhere (thinking about if we wanted to present users with autocomplete options on the client side when creating new environments).

If we do decide to stick with this approach we should definitely let users know that we are finding an org that matches and therefore assuming the user meant to specify `<org>/<project>` and if they meant project they should use the full qualifier

cmd/esc/cli/env.go
```diff
@@ -65,49 +66,197 @@ func newEnvCmd(esc *escCommand) *cobra.Command {
...
+		projectName:      projectName,
+		envName:          envName,
+		version:          version,
+		hasAmbiguousPath: hasAmbiguousPath,
+	}
+}
+
+// getEnvRef returns an environment reference corresponding to the given ref string
+// If a non-nil environmentRef is provided, default to its values if only a sole version is specified
+func (cmd *envCommand) getEnvRef(refString string, rel *environmentRef) (ref environmentRef) {
```
dschaller, line 132: super nit: 
```suggestion
func (cmd *envCommand) getEnvRef(refString string, rel *environmentRef) environmentRef {
```

cmd/esc/cli/env.go
```diff
@@ -65,49 +66,197 @@ func newEnvCmd(esc *escCommand) *cobra.Command {
...
+
+	return envRef
+}
+
+// Get an environment reference when creating a new environment
+// If the given path is ambiguous, we need to make additional API calls to disambiguate
+func (cmd *envCommand) getNewEnvRef(
+	ctx context.Context,
+	args []string,
+) (ref environmentRef, rest []string, err error) {
```
dschaller, line 149: For consistency with the rest of the codebase let's not name the return values for functions

cmd/esc/cli/env.go
dschaller: definitely not a blocker so feel free to continue without changing but I think `getExistingEnvRefWithFallback` might be clearer

cmd/esc/cli/cli_test.go
```diff
@@ -237,7 +237,8 @@ type testEnvironments struct {
 }
 
 func (e *testEnvironments) LoadEnvironment(ctx context.Context, envName string) ([]byte, eval.Decrypter, error) {
```
seanyeh, line 239: @pgavlin do you know why this method only receives environment name? The corresponding method in pulumi-service receives a ref string which includes `project/env` and evaluates correctly, but not sure what this test client is doing differently. maybe I'm misunderstanding something?

pgavlin: `envName` should include the entire ref string here. what behavior are you observing? have you attached a debugger and stepped through the failing tests?

dschaller (approved):
Generally looks good to me. Just for posterity we discussed offline there are two rare edge cases users might run into:
* users may specify a new project that is identical to their org name
* users may specify an org they don't belong not intending to create a new project

For the first one we can log that this conflict does exist and let users know we will be falling back to the legacy behavior and creating a new project in the default environment. If they want to create a new project they can do so using the full path. For the second we can't do much other than let them know that we couldn't find an org they belong to with that name and are therefore assuming they meant to create a new project in the default org. Both of these can be done in a follow up PR

seanyeh:
merging this into the feature branch

PR #376, opened 2024-09-05
[env] add clone command

dschaller:
add a new clone command for environments

```shell
> esc env clone localumi/clone/foobar foobar2 --preserve-history --preserve-access --preserve-env-tags --preserve-rev-tags
Environment localumi/default/foobar cloned into localumi/clone/foobar2.
```

### DIFF
--- a/cmd/esc/cli/env_clone.go
+++ b/cmd/esc/cli/env_clone.go
@@ -1,0 +1,97 @@
+// Copyright 2023, Pulumi Corporation.
+
+package cli
+
+import (
+	"context"
+	"fmt"
+	"strings"
+
+	"github.com/pulumi/esc/cmd/esc/cli/client"
+	"github.com/spf13/cobra"
+)
+
+func newEnvCloneCmd(env *envCommand) *cobra.Command {
+	var (
+		preserveHistory         bool
+		preserveAccess          bool
+		preserveEnvironmentTags bool
+		preserveRevisionTags    bool
+	)
+
+	cmd := &cobra.Command{
+		Use:   "clone [<org-name>/]<project-name>/<environment-name> [<project-name>/]<environment-name>",
+		Args:  cobra.MaximumNArgs(2),
+		Short: "Clone an existing environment into a new environment.",
+		Long: "Clone an existing environment into a new environment.\n" +
+			"\n" +
+			"This command clones an existing environment with the given identifier into a new environment.\n" +
+			"If a project is omitted from the new environment identifier the new environment will be created\n" +
+			"within the same project as the environment being cloned.\n",
+		SilenceUsage: true,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			ctx := context.Background()
+
+			if err := env.esc.getCachedClient(ctx); err != nil {
+				return err
+			}
+
+			ref, args, err := env.getNewEnvRef(ctx, args)
+			if err != nil {
+				return err
+			}
+			if ref.version != "" {
+				return fmt.Errorf("the clone command does not accept versions")
+			}
+
+			var destProject string
+			destName := args[0]
+			destParts := strings.Split(args[0], "/")
+			if len(destParts) == 2 {
+				destProject = destParts[0]
+				destName = destParts[1]
+			}
+
+			destEnv := client.CloneEnvironmentRequest{
+				Project:                 destProject,
+				Name:                    destName,
+				PreserveHistory:         preserveHistory,
+				PreserveAccess:          preserveAccess,
+				PreserveEnvironmentTags: preserveEnvironmentTags,
+				PreserveRevisionTags:    preserveRevisionTags,
+			}
+			if err := env.esc.client.CloneEnvironment(ctx, ref.orgName, ref.projectName, ref.envName, destEnv); err != nil {
+				return fmt.Errorf("cloning environment: %w", err)
+			}
+
+			if destProject == "" {
+				destProject = ref.projectName
+			}
+			fmt.Fprintf(
+				env.esc.stdout,
+				"Environment %s/%s/%s cloned into %s/%s/%s.\n",
+				ref.orgName, ref.projectName, ref.envName,
+				ref.orgName, destProject, destName,
+			)
+			return nil
+		},
+	}
+
+	cmd.Flags().BoolVar(&preserveHistory,
+		"history", false,
+		"preserve history of the environment being cloned")
+
+	cmd.Flags().BoolVar(&preserveAccess,
+		"access", false,
+		"add the newly cloned environment to the same teams that have access to the origin environment")
+
+	cmd.Flags().BoolVar(&preserveEnvironmentTags,
+		"envTags", false,
+		"preserve any tags on the environment being cloned")
+
+	cmd.Flags().BoolVar(&preserveRevisionTags,
+		"revTags", false,
+		"preserve any tags on the environment revisions being cloned")
+
+	return cmd
+}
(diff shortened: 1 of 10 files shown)


cmd/esc/cli/env_clone.go
```diff
@@ -0,0 +1,97 @@
+// Copyright 2023, Pulumi Corporation.
```
seanyeh, line 1: 2024

cmd/esc/cli/env_clone.go
```diff
@@ -0,0 +1,97 @@
...
+				"Environment %s/%s/%s cloned into %s/%s/%s.\n",
+				ref.orgName, ref.projectName, ref.envName,
+				ref.orgName, destProject, destName,
+			)
+			return nil
+		},
+	}
+
+	cmd.Flags().BoolVar(&preserveHistory,
+		"history", false,
```
seanyeh, line 81: would it be clearer if these were called `preserve-X` rather than just `X`? also I'm more used to seeing multiword flags in kebob case over camelcase, but if we already use camelcase elsewhere, it's fine to leave it here

cmd/esc/cli/env_clone.go
```diff
@@ -0,0 +1,97 @@
...
+		"preserve history of the environment being cloned")
+
+	cmd.Flags().BoolVar(&preserveAccess,
+		"access", false,
+		"add the newly cloned environment to the same teams that have access to the origin environment")
```
seanyeh, line 86: maybe "preserve the same team access on the environment being cloned"?

cmd/esc/cli/env_clone.go
```diff
@@ -0,0 +1,97 @@
...
+			if err != nil {
+				return err
+			}
+			if ref.version != "" {
+				return fmt.Errorf("the clone command does not accept versions")
+			}
+
+			var destProject string
+			destName := args[0]
+			destParts := strings.Split(args[0], "/")
```
seanyeh, line 49: I think it may be easier to use `Cut` here, and then check if the second part exists - if so, it's project/env otherwise just env.
Split works also, but then you'll need to return an error for len > 2

a part of me is also wondering if we could reuse the existing ref parsing methods but since we're not parsing org it'll be slightly different 🤷 probably fine to not reuse here

cmd/esc/cli/env_clone.go
```diff
@@ -0,0 +1,97 @@
...
+func newEnvCloneCmd(env *envCommand) *cobra.Command {
+	var (
+		preserveHistory         bool
+		preserveAccess          bool
+		preserveEnvironmentTags bool
+		preserveRevisionTags    bool
+	)
+
+	cmd := &cobra.Command{
+		Use:   "clone [<org-name>/]<project-name>/<environment-name> [<project-name>/]<environment-name>",
```
seanyeh, line 23: it'll be wordy but it would be nice to be clear which one is the old vs new. maybe `existing-project-name/existing-envionment-name` vs `new-project-name/new-environment-name...`, or `src vs dest` terminology

cmd/esc/cli/env_clone.go
```diff
@@ -0,0 +1,97 @@
...
+			"within the same project as the environment being cloned.\n",
+		SilenceUsage: true,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			ctx := context.Background()
+
+			if err := env.esc.getCachedClient(ctx); err != nil {
+				return err
+			}
+
+			ref, args, err := env.getNewEnvRef(ctx, args)
```
seanyeh, line 39: this is the source ref right? this should be `getExistingEnvRef`

seanyeh: now that i think about it, do we even want the legacy path defaulting (org/env)? If not, we might just want to call `getEnvRef` or `parseRef` directly

cmd/esc/cli/env_clone.go
dschaller: I was trying to keep these short since they already can't cleanly have a single character for the flags (e.g. `h`).

cmd/esc/cli/env_clone.go
dschaller: yeah I was thinking of reusing that function but a lot of that logic has things around optional org and project where as this will never have an org prefix so I wanted to remove the ability for that to be confused.

cmd/esc/cli/env_clone.go
dschaller: I'll add in src and dest

cmd/esc/cli/env_clone.go
dschaller: ah yeah good catch - I had to add a new field onto the environmentRef struct because I want to be able to explicitly fail if users are passing in just the environment name (omitting the project)

seanyeh: it's not pretty but I did a similar check like [this](https://github.com/pulumi/esc/blob/d3ce3545515bffea9656d58f4ae7bc0849bc3d40/cmd/esc/cli/env.go#L173) without using a field, would that work here? otherwise what you have here is also fine

seanyeh:
looks good to me besides [comment](https://github.com/pulumi/esc/pull/376#discussion_r1746412822) and fixing the tests

cmd/esc/cli/env_clone.go
dschaller: Hmm I could but it still wouldn't tell me if it's ambiguous because of the org or the project name since we would set the project to be the default value and it wouldn't be clear if we set it or the user did

cmd/esc/cli/env_clone.go
```diff
@@ -37,17 +37,23 @@ func newEnvCloneCmd(env *envCommand) *cobra.Command {
...
-			ref := env.parseRef(args[0])
-			if ref.isUsingLegacyID {
-				return errors.New("invalid source environment ID")
+			ref, args, err := env.getExistingEnvRef(ctx, args)
+			// An error will arise if an environment reference is ambiguous and can't be
+			// resolved to a single environment. The ref for the non-legacy environment will
+			// also be returned in this case.
+			// If the original ref is using a legacy ID of just env name return an error
+			// Otherwise we will ignore any conflict errors and assume the user meant <project-name>/<env-name>
+			if (err != nil && ref.isUsingLegacyID) || ref.isUsingLegacyID {
```
seanyeh, line 46: let's just simplify this to `if ref.isUsingLegacyID` (this is for both cases for err is nil or not), and make sure to add an `if err != nil` check after this conditional for the case where there is a non-legacy id related error

cmd/esc/cli/env_clone.go
```diff
@@ -37,17 +37,23 @@ func newEnvCloneCmd(env *envCommand) *cobra.Command {
...
+				return errors.New("source environment ID must be at least <project-name>/<env-name>")
```
seanyeh, line 47: "at least" isn't super clear. I would probably use something like what we have in the deprecation warning message (that you need either org/proj/env or proj/env)

cmd/esc/cli/env_clone.go
```diff
@@ -37,17 +37,23 @@ func newEnvCloneCmd(env *envCommand) *cobra.Command {
...
+
 			if ref.version != "" {
 				return errors.New("the clone command does not accept versions")
 			}
 
 			var destProject string
-			destName := args[1]
-			if project, name, hasDelimiter := strings.Cut(args[1], "/"); hasDelimiter {
+			destName := args[0]
+			if project, name, hasDelimiter := strings.Cut(args[0], "/"); hasDelimiter {
```
seanyeh, line 56: NP lets use destName instead of args[0] here

seanyeh: actually nvm, it doesn't really matter

seanyeh (approved):
spoke offline, lgtm once we add the custom error checking

PR #377, opened 2024-09-05
Add anonymous open environment routes

seanyeh:
Adds methods for the anonymous open env routes so that the Pulumi CLI can use them

### DIFF
--- a/cmd/esc/cli/client/client.go
+++ b/cmd/esc/cli/client/client.go
@@ -180,6 +180,10 @@ type Client interface {
 	// environment envName and org orgName.
 	GetOpenEnvironmentWithProject(ctx context.Context, orgName, projectName, envName, openEnvID string) (*esc.Environment, error)
 
+	// GetAnonymousOpenEnvironment returns the AST, values, and schema for the open environment with ID openEnvID in
+	// an anonymous environment.
+	GetAnonymousOpenEnvironment(ctx context.Context, orgName, openEnvID string) (*esc.Environment, error)
+
 	// GetOpenProperty returns the value of a single property in the open environment with ID openEnvID in
 	// environment envName and org orgName.
 	//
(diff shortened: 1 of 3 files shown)


cmd/esc/cli/client/client.go
```diff
@@ -677,6 +685,16 @@ func (pc *client) GetOpenEnvironmentWithProject(ctx context.Context, orgName, pr
 	return &resp, nil
 }
 
+func (pc *client) GetAnonymousOpenEnvironment(ctx context.Context, orgName, openSessionID string) (*esc.Environment, error) {
+	var resp esc.Environment
+	path := fmt.Sprintf("/api/esc/environments/%v/yaml/open/%v", orgName, openSessionID)
```
IaroslavTitov, line 690: Should this include project in the URL? As far as I understand, envName and project are not used, but eventually we will remove routes without project, right?

Also, nit - ideally should be more specific and use %s for strings, generic is always more bug-prone

cmd/esc/cli/client/client.go
```diff
@@ -693,6 +711,22 @@ func (pc *client) GetOpenProperty(ctx context.Context, orgName, projectName, env
...
 
+func (pc *client) GetAnonymousOpenProperty(ctx context.Context, orgName, openSessionID, property string) (*esc.Value, error) {
+	queryObj := struct {
+		Property string `url:"property"`
+	}{
+		Property: property,
+	}
+
+	var resp esc.Value
+	path := fmt.Sprintf("/api/esc/environments/%v/yaml/open/%v", orgName, openSessionID)
```
IaroslavTitov, line 722: Same as above

cmd/esc/cli/client/client.go
seanyeh: we will remove preview routes but this is a GA route. we decided that we don't need to include projects in this route for anonymous envs.

IaroslavTitov (approved):

PR #379, opened 2024-09-09
Update pu/pu dependency and use diy rather than filestate

Frassle:
`filestate.IsFileStateBackendURL` was replaced with `diy.IsDIYBackendURL` but we've had to leave the old filestate one around because of cyclic dependencies with ESC.

This updates the esc codebase to get rid of that usage so that we can then update pu/pu and clean up this old method.


dschaller:
Can we update the changelog to communicate that we are changing the minimum go version to 1.21?

dschaller (approved):

PR #382, opened 2024-09-10
[env] add project filter to ls command

dschaller:
```bash
$ esc env ls -p {project}
```

### DIFF
--- a/cmd/esc/cli/env_ls.go
+++ b/cmd/esc/cli/env_ls.go
@@ -13,7 +13,10 @@ import (
 )
 
 func newEnvLsCmd(env *envCommand) *cobra.Command {
-	var orgFilter string
+	var (
+		orgFilter     string
+		projectFilter string
+	)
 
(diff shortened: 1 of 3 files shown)


cmd/esc/cli/env_ls.go
```diff
@@ -56,11 +59,13 @@ func newEnvLsCmd(env *envCommand) *cobra.Command {
 
 	cmd.PersistentFlags().StringVarP(
 		&orgFilter, "organization", "o", "", "Filter returned stacks to those in a specific organization")
+	cmd.PersistentFlags().StringVarP(
+		&projectFilter, "project", "p", "", "Filter returned stacks to those in a specific project")
```
seanyeh, line 63: I know this is just copy from the line above, but this really should be `environments` instead of `stacks` right? (it's probably copypasta from pulumi/pulumi)

seanyeh (approved):

PR #384, opened 2024-09-11
chore: freeze 0.10.0

seanyeh:


komalali:
Do we need to update the changelog pending and whatnot?

pgavlin:
that's here I think: https://github.com/pulumi/esc/pull/385

dschaller:
Yeah seems like we generally land the changelog updates and then freeze the version

komalali (approved):

PR #394, opened 2024-09-18
Make the value arg optional for set command

### LINKED ISSUE #340: Add option to prompt and mask value when using `--secret` for `pulumi env`
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details

`pulumi env set test key --secret` expects the secret value to be passed as an argument. This is not optimal, as it's saved in the terminal history as plain-text, etc. Instead, consider adding a prompt/mask experience similar to that of `pulumi config set key --secret`.

praneetloke:
Fixes #340.

This PR uses the same code as Pulumi CLI to prompt the user for a value if no value is specificed when in interactive mode.

Although I updated the test, I couldn't get it to pass because the pipe command seems to not run correctly? I couldn't quite figure out what is going on and hope that one of you knows how to get the test to pass. FWIW, I built a local version of the CLI and have confirmed it to work with these changes, i.e. I can run `echo 'secret' | esc env set <workspace> <path> --secret` (as well as without `--secret`.)

### DIFF
--- a/cmd/esc/cli/testdata/env-set-secret.yaml
+++ b/cmd/esc/cli/testdata/env-set-secret.yaml
@@ -16,6 +16,8 @@ run: |
   esc env get default/test
   esc env set default/test password true --secret
   esc env get default/test
+  echo 'mysecret' | esc env set default/test password --secret
+  esc env get default/test --show-secrets
   esc env set default/test password '[]' --secret || echo OK
 stdout: |
   > esc env init default/test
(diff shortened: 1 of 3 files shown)


cmd/esc/cli/testdata/env-set-secret.yaml
```diff
@@ -16,6 +16,8 @@ run: |
   esc env get default/test
   esc env set default/test password true --secret
   esc env get default/test
+  echo 'mysecret' | esc env set default/test password --secret
```
praneetloke, line 19: As mentioned in the PR description, I can't get this to work. It's as if `echo 'mysecret'` evaluates to an empty string because that's what the value of `esc env set ...` ends up getting. However, building a local version of the CLI and executing the same thing works as expected, so I am not sure what is going on with the test runner.

PR #397, opened 2024-09-23
fix installation instructions for pulumi esc

tgummerer:
Just using curl get.pulumi.com will install the pulumi CLI.  While that embeds esc as `pulumi env` it's not what we want to direct users to when they look at esc, especially as the guide linked above refers to plain `esc`.


seanyeh:
looks like this is a dupe of https://github.com/pulumi/esc/pull/374 :)

tgummerer:
Oh whoops, I didn't realize that.  Let's merge that one since it's the exact same anyway.

PR #410, opened 2024-10-26
[cli] proxy authentication from new CLI options

dschaller:
At the moment `pulumi env ls` will not respect the cloud backend that is set within the pulumi CLI. For example, if you run a `pulumi login https://localhost:8080` and a subsequent `pulumi env ls` you will see all the environments listed in the production Pulumi cloud.

This PR modifies that behavior to set the esc cli account to the credentials passed into the esc cli from the pulumi cli

Screenshot 2024-10-26 at 00 13 56

### DIFF
--- a/cmd/esc/cli/esc.go
+++ b/cmd/esc/cli/esc.go
@@ -85,6 +85,31 @@ func newESC(opts *Options) *escCommand {
 		esc.newClient = client.New
 	}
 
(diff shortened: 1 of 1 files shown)


dschaller:
Turns out this is working and I must have confused my two pulumi binaries when logging into different accounts 😬

PR #411, opened 2024-10-28
[syntax] Refactor YAML editing functionality

pgavlin:
Replace the CLI's `yamlNode` type and methods with equivalents on `syntax.YAMLNode`.

This is a bit duplicative of functionality already in the syntax package, which allows consumers to edit syntax trees without knowledge of the concrete syntax. However, this is the lowest-lift option for refactoring this functionality with high confidence in the results. We should replace the `YAMLNode` methods with higher-level methods that are concrete-syntax-agnostic in the future.


seanyeh (approved):
lgtm 👍

Issue #415, opened 2024-11-05
Add command to print out raw definition


seanyeh:
## Hello!
- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details

Currently, `esc env get <env>` will print out a "value" section with the values and "definition" with the raw definition. It would be useful to have a command that only returns the raw definition

PR #416, opened 2024-11-05
Add --definition to get to print definition

seanyeh:
Fixes https://github.com/pulumi/esc/issues/415

Adds a `--definition` flag to the get command. we already have a `--value` command to only output the value, so I figured the `--definition` flag makes sense. open to other suggestions though

## Testing

```
❯ go run ./cmd/esc env get temp2/abc/importer --definition --value json
Error: `--value` and `--definition` flags cannot be used together
exit status 1

❯ go run ./cmd/esc env get temp2/abc/importer --definition
imports:
  - abc/envvars
...
```

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -1,5 +1,8 @@
 ### Improvements
 
+- Add `--definition` flag to `esc env get` to output definition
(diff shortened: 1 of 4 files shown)


komalali (approved):

komalali:
The `--definition` flag makes sense to me!

Issue #370, opened 2024-08-13
UpdateEnvironmentWithRevision does not return diagnostics properly


IaroslavTitov:
### What happened?

Method `UpdateEnvironmentWithRevision` does not return diagnostics properly. Instead, empty array is returned and err is returned, filled with diagnostics in the error text.

### Example

Call like this:
```
diagnostics, revision, err := st.client.UpdateEnvironmentWithRevision(context.Background(), input.OrgName, input.EnvName, input.Yaml, "")
	if diagnostics != nil {
		return nil, fmt.Errorf("failed to update environment, yaml code failed following checks: %+v", diagnostics)
	}
	if err != nil {
		return nil, fmt.Errorf("failed to update environment due to error: %+v", err)
	}
```

Diagnostics check is skipped, because it's empty array, then second if is hit. The error returned is 
`failed to update environment due to error: [0]
    Diags: string must match the pattern "^[a-zA-Z0-9-_.]{1,100}\\/[a-zA-Z0-9-_.]{1,100}$"`

Which shows that diags are present, just not parsed and returned correctly.

### Output of `pulumi about`

Not relevant

### Additional context

Found this in Pulumi Service Provider

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

PR #422, opened 2024-12-09
Fix diagnostic messages on update environment

seanyeh:
Fixes https://github.com/pulumi/esc/issues/370

When updating an env with an invalid definition, `UpdateEnvironmentWithRevision` should return no error but with diagnostics

## Testing

Because of this issue, we previously wouldn't reach [here](https://github.com/pulumi/esc/blob/0a8fc4100f4cce21f9b45faa9ef379d100718696/cmd/esc/cli/env_edit.go#L127) (which requires non-error but with diagnostics), but with this change, now we can, because `UpdateEnvironmentWithRevision` will now return no error but with diagnostics on invalid definitions

When updating an env with an invalid definition:

Before:
```
❯ go run ./cmd/esc env edit test3/abc/ghi
Error: updating environment definition: [0]
Diags: yaml: line 2: mapping values are not allowed in this context

exit status 1
```

Now:
```
❯ go run ./cmd/esc env edit test3/abc/ghi
Error: yaml: line 2: mapping values are not allowed in this context

Press ENTER to continue editing or ^D to exit
```

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -1,7 +1,7 @@
 ### Improvements
 
-- Add `--definition` flag to `esc env get` to output definition
-  [#416](https://github.com/pulumi/esc/pull/416)
+- Fix diagnostic messages when updating environment with invalid definition
+  [#422](https://github.com/pulumi/esc/pull/422)
 
 ### Bug Fixes
 
--- a/cmd/esc/cli/client/client.go
+++ b/cmd/esc/cli/client/client.go
@@ -560,7 +560,7 @@ func (pc *client) UpdateEnvironmentWithRevision(
 	})
 	if err != nil {
 		var diags *EnvironmentErrorResponse
-		if errors.As(err, &diags) && diags.Code == http.StatusBadRequest && len(diags.Diagnostics) != 0 {
+		if errors.As(err, &diags) && len(diags.Diagnostics) != 0 {
 			return diags.Diagnostics, 0, nil
 		}
 		return nil, 0, err
--- a/cmd/esc/cli/client/client_test.go
+++ b/cmd/esc/cli/client/client_test.go
@@ -339,6 +339,39 @@ func TestUpdateEnvironment(t *testing.T) {
 		assert.Equal(t, expected, diags)
 	})
 
+	t.Run("Diags only", func(t *testing.T) {
+		expected := []EnvironmentDiagnostic{
+			{
+				Range: &esc.Range{
+					Environment: "test-env",
+					Begin:       esc.Pos{Line: 42, Column: 1},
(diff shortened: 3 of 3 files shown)


cmd/esc/cli/client/client.go
```diff
@@ -560,7 +560,7 @@ func (pc *client) UpdateEnvironmentWithRevision(
 	})
 	if err != nil {
 		var diags *EnvironmentErrorResponse
-		if errors.As(err, &diags) && diags.Code == http.StatusBadRequest && len(diags.Diagnostics) != 0 {
+		if errors.As(err, &diags) && (diags.Code == http.StatusBadRequest || len(diags.Diagnostics) != 0) {
```
seanyeh, line 563: This is because on certain 4xx errors, the pulumi service returns an object without {..., `"code": 4xx`}

CHANGELOG_PENDING.md
```diff
@@ -1,7 +1,7 @@
 ### Improvements
 
-- Add `--definition` flag to `esc env get` to output definition
-  [#416](https://github.com/pulumi/esc/pull/416)
```
IaroslavTitov, line 4: Did you mean to overwrite this previous entry in Improvements?

cmd/esc/cli/client/client.go
IaroslavTitov: Hmm, should we fix this in PS? Sounds like it's not following a required API contract?

cmd/esc/cli/client/client_test.go
```diff
@@ -326,8 +326,6 @@ func TestUpdateEnvironment(t *testing.T) {
 			w.WriteHeader(http.StatusBadRequest)
 
 			err := json.NewEncoder(w).Encode(EnvironmentErrorResponse{
-				Code:        400,
-				Message:     "bad request",
```
IaroslavTitov, line 330: Could we keep both test paths, like with and without `Code`?

CHANGELOG_PENDING.md
seanyeh: yup this was from the previous release

cmd/esc/cli/client/client.go
seanyeh: I initially considered this but I think it's a much bigger lift - we follow this "pattern" in many routes where certain errors will return `code` and `message` by default, but otherwise not.

I think perhaps a better fix is to check the http response status code instead. @pgavlin I notice that we [don't return](https://github.com/pulumi/esc/blob/0a8fc4100f4cce21f9b45faa9ef379d100718696/cmd/esc/cli/client/client.go#L1156-L1174) the http response on 4xx/5xx errors, is this intentional?

IaroslavTitov (approved):

PR #432, opened 2025-01-15
provider rotation

nyobe:
Introduces a new provider type and verb to support rotating static credentials.

Rotator providers have an additional `state` input which is used as a write-back target. The result of invoking Rotate is persisted back into the environment to this key.  This provides a stable location to write to without accidentally clobbering interpolations that might be used for other inputs.

The new `fn::rotate` function type is used to invoke this new type of provider. It behaves similarly to existing providers during Open, but during rotation the evaluator will invoke the Rotate methods and collect their outputs, which are returned to the caller. The caller is expected to persist these updates back into the environment as a new revision.

```
# long form
fn::rotate:
  rotator: provider
  inputs:
    # any config the provider needs to use during rotation
  state:
    # persisted state

# short form
fn::rotate::provider:
  inputs:
    # any config the provider needs to use during rotation
  state:
    # persisted state
```
  

Closes https://github.com/pulumi/pulumi-service/issues/24986

### DIFF
--- a/ast/expr.go
+++ b/ast/expr.go
@@ -439,32 +439,32 @@ func Open(provider string, inputs *ObjectExpr) *OpenExpr {
 type RotateExpr struct {
 	builtinNode
 
-	Provider *StringExpr
-	Inputs   Expr
-	State    *ObjectExpr
+	Rotator *StringExpr
+	Inputs  Expr
+	State   *ObjectExpr
 }
 
-func RotateSyntax(node *syntax.ObjectNode, name *StringExpr, args Expr, provider *StringExpr, inputs Expr, state *ObjectExpr) *RotateExpr {
+func RotateSyntax(node *syntax.ObjectNode, name *StringExpr, args Expr, rotator *StringExpr, inputs Expr, state *ObjectExpr) *RotateExpr {
 	return &RotateExpr{
 		builtinNode: builtin(node, name, args),
-		Provider:    provider,
+		Rotator:     rotator,
 		Inputs:      inputs,
 		State:       state,
 	}
 }
 
-func Rotate(provider string, inputs, state *ObjectExpr) *RotateExpr {
-	name, providerX := String("fn::rotate"), String(provider)
+func Rotate(rotator string, inputs, state *ObjectExpr) *RotateExpr {
+	name, rotatorX := String("fn::rotate"), String(rotator)
 
 	entries := []ObjectProperty{
-		{Key: String("provider"), Value: providerX},
+		{Key: String("rotator"), Value: rotatorX},
 		{Key: String("inputs"), Value: inputs},
 		{Key: String("state"), Value: state},
 	}
 
 	return &RotateExpr{
 		builtinNode: builtin(nil, name, Object(entries...)),
-		Provider:    providerX,
+		Rotator:     rotatorX,
 		Inputs:      inputs,
 		State:       state,
 	}
@@ -745,15 +745,15 @@ func parseRotate(node *syntax.ObjectNode, name *StringExpr, args Expr) (Expr, sy
 		return RotateSyntax(node, name, args, nil, nil, nil), diags
 	}
 
-	var providerExpr, inputs, stateExpr Expr
+	var rotatorExpr, inputs, stateExpr Expr
 	var diags syntax.Diagnostics
 
 	for i := 0; i < len(obj.Entries); i++ {
 		kvp := obj.Entries[i]
 		key := kvp.Key
 		switch key.GetValue() {
-		case "provider":
-			providerExpr = kvp.Value
+		case "rotator":
+			rotatorExpr = kvp.Value
 		case "inputs":
 			inputs = kvp.Value
 		case "state":
@@ -761,17 +761,17 @@ func parseRotate(node *syntax.ObjectNode, name *StringExpr, args Expr) (Expr, sy
 		}
 	}
 
-	provider, ok := providerExpr.(*StringExpr)
+	provider, ok := rotatorExpr.(*StringExpr)
 	if !ok {
-		if providerExpr == nil {
-			diags.Extend(ExprError(obj, "missing provider name ('provider')"))
+		if rotatorExpr == nil {
+			diags.Extend(ExprError(obj, "missing rotator name ('rotator')"))
 		} else {
-			diags.Extend(ExprError(providerExpr, "provider name must be a string literal"))
+			diags.Extend(ExprError(rotatorExpr, "rotator name must be a string literal"))
 		}
 	}
 
 	if inputs == nil {
-		diags.Extend(ExprError(obj, "missing provider inputs ('inputs')"))
+		diags.Extend(ExprError(obj, "missing rotator inputs ('inputs')"))
 	}
 
 	state, ok := stateExpr.(*ObjectExpr)
--- a/ast/expr.go
+++ b/ast/expr.go
@@ -18,6 +18,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"reflect"
+	"slices"
 	"strings"
 
 	"github.com/hashicorp/hcl/v2"
@@ -436,12 +437,36 @@ func Open(provider string, inputs *ObjectExpr) *OpenExpr {
 
 // RotateExpr is a type of OpenExpr that supports a rotate operation.
 type RotateExpr struct {
-	*OpenExpr
+	builtinNode
+
+	Provider *StringExpr
+	Inputs   Expr
+	State    *ObjectExpr
 }
 
-func RotateSyntax(node *syntax.ObjectNode, name *StringExpr, args Expr, provider *StringExpr, inputs Expr) *RotateExpr {
+func RotateSyntax(node *syntax.ObjectNode, name *StringExpr, args Expr, provider *StringExpr, inputs Expr, state *ObjectExpr) *RotateExpr {
 	return &RotateExpr{
-		OpenExpr: OpenSyntax(node, name, args, provider, inputs),
+		builtinNode: builtin(node, name, args),
+		Provider:    provider,
+		Inputs:      inputs,
+		State:       state,
+	}
+}
+
+func Rotate(provider string, inputs, state *ObjectExpr) *RotateExpr {
+	name, providerX := String("fn::rotate"), String(provider)
+
+	entries := []ObjectProperty{
+		{Key: String("provider"), Value: providerX},
+		{Key: String("inputs"), Value: inputs},
+		{Key: String("state"), Value: state},
+	}
+
+	return &RotateExpr{
+		builtinNode: builtin(nil, name, Object(entries...)),
+		Provider:    providerX,
+		Inputs:      inputs,
+		State:       state,
 	}
 }
 
@@ -716,11 +741,11 @@ func parseShortOpen(node *syntax.ObjectNode, name *StringExpr, args Expr) (Expr,
 func parseRotate(node *syntax.ObjectNode, name *StringExpr, args Expr) (Expr, syntax.Diagnostics) {
 	obj, ok := args.(*ObjectExpr)
 	if !ok {
-		diags := syntax.Diagnostics{ExprError(args, "the argument to fn::rotate must be an object containing 'provider' and 'inputs'")}
-		return RotateSyntax(node, name, args, nil, nil), diags
+		diags := syntax.Diagnostics{ExprError(args, "the argument to fn::rotate must be an object containing 'provider', 'inputs' and 'state'")}
+		return RotateSyntax(node, name, args, nil, nil, nil), diags
 	}
 
-	var providerExpr, inputs Expr
+	var providerExpr, inputs, stateExpr Expr
 	var diags syntax.Diagnostics
 
 	for i := 0; i < len(obj.Entries); i++ {
@@ -731,6 +756,8 @@ func parseRotate(node *syntax.ObjectNode, name *StringExpr, args Expr) (Expr, sy
 			providerExpr = kvp.Value
 		case "inputs":
 			inputs = kvp.Value
+		case "state":
+			stateExpr = kvp.Value
 		}
 	}
 
@@ -747,18 +774,40 @@ func parseRotate(node *syntax.ObjectNode, name *StringExpr, args Expr) (Expr, sy
 		diags.Extend(ExprError(obj, "missing provider inputs ('inputs')"))
 	}
 
-	return RotateSyntax(node, name, obj, provider, inputs), diags
+	state, ok := stateExpr.(*ObjectExpr)
+	if !ok && state != nil {
+		diags.Extend(ExprError(stateExpr, "rotation state must be an object literal"))
+	}
+
+	return RotateSyntax(node, name, obj, provider, inputs, state), diags
 }
 
 func parseShortRotate(node *syntax.ObjectNode, name *StringExpr, args Expr) (Expr, syntax.Diagnostics) {
 	kvp := node.Index(0)
 	provider := StringSyntaxValue(name.Syntax().(*syntax.StringNode), strings.TrimPrefix(kvp.Key.Value(), "fn::rotate::"))
-	if args == nil {
-		diags := syntax.Diagnostics{ExprError(name, "missing provider inputs")}
-		return RotateSyntax(node, name, args, provider, nil), diags
+
+	inputs, ok := args.(*ObjectExpr)
+	if !ok {
+		diags := syntax.Diagnostics{ExprError(args, "provider inputs must be an object")}
+		return RotateSyntax(node, name, args, nil, nil, nil), diags
 	}
 
-	return RotateSyntax(node, name, args, provider, args), nil
+	// hoist 'state' key out of inputs
+	var stateExpr Expr
+	if i := slices.IndexFunc(inputs.Entries, func(kvp ObjectProperty) bool {
+		return kvp.Key.GetValue() == "state"
+	}); i != -1 {
+		stateExpr = inputs.Entries[i].Value
+		inputs.Entries = slices.Delete(inputs.Entries, i, i+1)
+	}
+
+	state, ok := stateExpr.(*ObjectExpr)
+	if !ok && state != nil {
+		diags := syntax.Diagnostics{ExprError(stateExpr, "rotation state must be an object literal")}
+		return RotateSyntax(node, name, args, nil, nil, nil), diags
+	}
+
+	return RotateSyntax(node, name, args, provider, inputs, state), nil
 }
 
 func parseJoin(node *syntax.ObjectNode, name *StringExpr, args Expr) (Expr, syntax.Diagnostics) {
--- a/ast/expr.go
+++ b/ast/expr.go
@@ -438,32 +438,32 @@ func Open(provider string, inputs *ObjectExpr) *OpenExpr {
 type RotateExpr struct {
 	builtinNode
 
-	Rotator *StringExpr
-	Inputs  Expr
-	State   *ObjectExpr
+	Provider *StringExpr
+	Inputs   Expr
+	State    *ObjectExpr
 }
 
-func RotateSyntax(node *syntax.ObjectNode, name *StringExpr, args Expr, rotator *StringExpr, inputs Expr, state *ObjectExpr) *RotateExpr {
+func RotateSyntax(node *syntax.ObjectNode, name *StringExpr, args Expr, provider *StringExpr, inputs Expr, state *ObjectExpr) *RotateExpr {
 	return &RotateExpr{
 		builtinNode: builtin(node, name, args),
-		Rotator:     rotator,
+		Provider:    provider,
 		Inputs:      inputs,
 		State:       state,
 	}
 }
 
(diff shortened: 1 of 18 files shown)


eval/patch.go
```diff
@@ -0,0 +1,78 @@
...
+
+// Patch represents a value that should be written back to the environment at the given path.
+type Patch struct {
+	DocPath     string
+	Replacement esc.Value
+}
+
+// ApplyPatches applies a set of patches values to an environment definition.
+// If patch values contain secret values, they will be wrapped with fn::secret.
+func ApplyPatches(source []byte, patches []*Patch) ([]byte, error) {
```
dschaller, line 18: Not blocking just more of a thought - I wonder if we would ever want to have this be generic enough to apply any patch instead of limiting this to just values. I know for our current use case this isn't in scope but if a customer wants to use this to apply a patch for automating a migration of which requires import changes it could be helpful.

If we stick with just values here we may want to capture that in the function name

nyobe: that's fair, it could drop the hardcoded assumption of the `values` toplevel key.
the name is also a bit of a misnomer since it's not doing a json patch, it's just doing a replacement.  Perhaps `BulkSetValues` or `ApplyReplacements` or something like that? idk

provider.go
```diff
@@ -29,3 +29,16 @@ type Provider interface {
...
 }
+
+// A Rotator enables environments to rotate a secret.
+// Rotators persist internal state by writing back to the environment.
+type Rotator interface {
+	// Schema returns the provider's input, state, and output schemata.
+	Schema() (inputs, state, outputs *schema.Schema)
+
+	// Open retrieves the provider's secrets.
+	Open(ctx context.Context, inputs, state map[string]Value, executionContext EnvExecContext) (Value, error)
```
nyobe, line 40: possibly Open should only take `state`, not `inputs`, because longer term we don't want to necessarily evaluate managing credential inputs during Open, just during Rotate.

eval/eval.go
```diff
@@ -98,7 +105,28 @@ func CheckEnvironment(
 	execContext *esc.ExecContext,
 	showSecrets bool,
 ) (*esc.Environment, syntax.Diagnostics) {
-	return evalEnvironment(ctx, true, name, env, decrypter, providers, environments, execContext, showSecrets)
+	checked, _, diags := evalEnvironment(ctx, true, name, env, decrypter, providers, nil, environments, execContext, showSecrets, nil)
```
seanyeh, line 108: I think we'll need to accept a rotatorLoader in CheckEnvironment to pass to the eval

nyobe: you are totally right 🤦‍♀️

eval/eval.go
```diff
@@ -932,6 +985,86 @@ func (e *evalContext) evaluateBuiltinOpen(x *expr, repr *openExpr) *value {
...
+			e.errorf(repr.syntax(), "internal error: invalid schema (%v)", err)
+		} else {
+			x.schema = outputSchema
+		}
+	}
+	v.schema = x.schema
+
+	inputs, inputsOK := e.evaluateTypedExpr(repr.inputs, repr.inputSchema)
+	state, stateOK := e.evaluateTypedExpr(repr.state, repr.stateSchema)
+	if !inputsOK || inputs.containsUnknowns() || !stateOK || state.containsUnknowns() || e.validating || e.rotatePaths == nil || err != nil {
```
seanyeh, line 1024: when I open an environment, the rotator `Open` doesn't get called because it hits this conditional (e.rotatePaths is nil). I think when e.rotatePaths is nil, we should still call rotator.Open?

provider.go
dschaller, line 40: yeah I would drop the inputs from open since we don't really need the inputs to perform the open. Also, it encourages best practice of not having the inputs be considered part of the open permissions

eval/eval.go
```diff
@@ -109,15 +139,17 @@ func evalEnvironment(
 	env *ast.EnvironmentDecl,
 	decrypter Decrypter,
 	providers ProviderLoader,
+	rotators RotatorLoader,
 	envs EnvironmentLoader,
 	execContext *esc.ExecContext,
 	showSecrets bool,
-) (*esc.Environment, syntax.Diagnostics) {
+	rotatePaths map[string]bool,
```
dschaller, line 146: @pgavlin wdyt about removing this? is there a strong use case for rotating specific paths within an environment?

eval/eval.go
```diff
@@ -165,6 +198,9 @@ type evalContext struct {
 	root      *expr  // the root expression
 	base      *value // the base value
 
+	rotatePaths  map[string]bool // when non-nil, specifies providers to rotate. if empty, the full environment is rotated.
```
nyobe, line 201: this might be too overloaded, maybe we should have an explicit `rotating` boolean instead of a nil check

eval/patch.go
```diff
@@ -0,0 +1,92 @@
...
+			for k, v := range pv {
+				m[k] = valueToSecretJSON(v)
+			}
+			return m
+		default:
+			return pv
+		}
+	}()
+	// wrap secret values
+	if v.Secret {
```
dschaller, line 86: Would this only wrap the top most value as a secret instead of the potentially nested values that are secrets? (I could be wrong about this but adding a test should help determine this)

provider.go
```diff
@@ -29,3 +29,16 @@ type Provider interface {
 	// Open retrieves the provider's secrets.
 	Open(ctx context.Context, inputs map[string]Value, executionContext EnvExecContext) (Value, error)
 }
+
+// A Rotator enables environments to rotate a secret.
+// Rotators persist internal state by writing back to the environment.
+type Rotator interface {
+	// Schema returns the provider's input, state, and output schemata.
+	Schema() (inputs, state, outputs *schema.Schema)
```
nyobe, line 37: We talked about it being confusing that `open` doesn't necessarily return the full state, so perhaps this can be simplified so that the `output` schema is used for `state` schema?

(This would also help heal the divergence of the Rotator and Provider interfaces)

eval/patch.go
nyobe: https://github.com/pulumi/esc/pull/432/commits/14d609a2e418cc634fa50fb965449b96e10fa2a0

provider.go
pgavlin: I would not drop the inputs. It might be that the provider still needs creds to fetch secrets (e.g. if the secrets aren't embedded as literals in the state, and are instead references).

provider.go
pgavlin: It might be a little unclear, but I'd suggest merging the `Rotator` and `Provider` interfaces a bit more:

```
type Provider interface {
	// Schema returns the provider's input, state, and output schemata.
	Schema() (inputs, state, outputs *schema.Schema)

	// Open retrieves the provider's secrets.
	Open(ctx context.Context, inputs map[string]Value, state map[string]Value, executionContext EnvExecContext) (Value, error)
}

type RotatableProvider interface {
	// Rotate rotates the provider's secret, and returns the rotator's new state to be persisted.
	Rotate(ctx context.Context, inputs, state map[string]Value) (map[string]Value, error)
}
```

eval/eval.go
```diff
@@ -39,6 +39,12 @@ type ProviderLoader interface {
 	LoadProvider(ctx context.Context, name string) (esc.Provider, error)
 }
 
+// A RotatorLoader provides the environment evaluator the capability to load rotators.
+type RotatorLoader interface {
+	// LoadRotator loads the rotator with the given name.
+	LoadRotator(ctx context.Context, name string) (esc.Rotator, error)
+}
+
```
pgavlin, line 47: I don't think that this should be a separate interface.

I think that we should use the same `Loader` interface and check the result of `LoadProvider` as needed using a type assertion.

eval/eval.go
pgavlin: Unclear. I don't have strong feelings either way about keeping this or not. We can always add it later--it's not a super-tricky feature.

eval/eval.go
nyobe: Yeah, this was only necessary due to the interfaces diverging

eval/eval.go
nyobe: it's mainly there to support [an inline rotation button in the table view](https://docs.google.com/presentation/d/1_NC_QM31eBaVghlAUGyb1fQj0C_zXdF5LZ14elGOWB8/edit#slide=id.g32535248952_2_38)

ast/expr.go
```diff
@@ -696,6 +738,78 @@ func parseShortOpen(node *syntax.ObjectNode, name *StringExpr, args Expr) (Expr,
...
+			diags.Extend(ExprError(providerExpr, "provider name must be a string literal"))
+		}
+	}
+
+	if inputs == nil {
+		diags.Extend(ExprError(obj, "missing provider inputs ('inputs')"))
+	}
+
+	state, ok := stateExpr.(*ObjectExpr)
+	if !ok && state != nil {
```
seanyeh, line 778: for some reason i'm getting a lint error here locally `impossible condition: nil != nil` but I'm not sure why 🤔

eval/expr.go
```diff
@@ -204,6 +205,40 @@ func (x *expr) export(environment string) esc.Expr {
...
+				}).Schema(),
+				Arg: esc.Expr{
+					Object: map[string]esc.Expr{
+						"provider": repr.provider.export(environment),
+						"inputs":   repr.inputs.export(environment),
+						"state":    repr.state.export(environment),
+					},
+				},
+			}
+		} else {
```
seanyeh, line 227: just so i'm following, this is for when the shorthand form is used?

seanyeh (commented):
looks good to me 🔥

dschaller:
> I would not drop the inputs. It might be that the provider still needs creds to fetch secrets (e.g. if the secrets aren't embedded as literals in the state, and are instead references).

That's a good point - it is possible state isn't stored here so we might need credentials to open the rotated credentials.

eval/eval.go
```diff
@@ -98,26 +99,48 @@ func CheckEnvironment(
...
+		rotatePaths[path] = true
+	}
+	return evalEnvironment(ctx, false, true, name, env, decrypter, providers, environments, execContext, true, rotatePaths)
 }
 
 // evalEnvironment evaluates an environment and exports the result of evaluation.
 func evalEnvironment(
 	ctx context.Context,
 	validating bool,
+	rotating bool,
```
dschaller, line 129: Based on todays discussion I would probably have this be `RotationContext` that has a boolean to say if a rotation is occurring to make expanding this easier if we want to add paths later to the context

ast/expr.go
nyobe: hmm, do you see the same error for the `providerExpr.(*StringExpr)` above?

eval/expr.go
nyobe: yes, that's right, in which case we need to construct a synthetic argument combining inputs & state, which were split during parsing

eval/eval.go
nyobe: based on sync today, dropping support for this for now https://github.com/pulumi/esc/pull/432/commits/a208a03e6a11b74558214072ce8da3a29f841017

provider.go
nyobe: I feel a little bit unsure about this, I think that if we don't drop inputs during open then there isn't as strong an argument for splitting state out of inputs at all 🤔

eval/eval.go
nyobe: since this is unexported, that doesn't feel super necessary, unless you're suggesting that the RotateEnvironment api should take a RotationContext argument?

dschaller: Ah yeah sorry that was more my train of thought. Internal is not a big deal but the `RotateEnvironment` function above which calls this

ast/expr.go
```diff
@@ -434,6 +435,41 @@ func Open(provider string, inputs *ObjectExpr) *OpenExpr {
...
+		Inputs:      inputs,
+		State:       state,
+	}
+}
+
+func Rotate(provider string, inputs, state *ObjectExpr) *RotateExpr {
+	name, providerX := String("fn::rotate"), String(provider)
+
+	entries := []ObjectProperty{
+		{Key: String("provider"), Value: providerX},
```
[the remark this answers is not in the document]
nyobe: hmm, yeah maybe, this is a holdover from extending `fn::open`, but rotators are a subset of providers

provider.go
nyobe: as we discussed, extended the ProviderLoader interface to support both LoadProvider and LoadRotator, so the interfaces can be separate

provider.go
pgavlin: > part of the reason for separating them is to be able to not evaluate managing creds during open

I don't think that dropping the inputs is the way we'll achieve this. I think we'll get that done through new semantics around managing creds. Dropping the inputs effectively forces all rotatable providers to be fully static at time of open, which I don't think is what we want.

nyobe: Ok! Now that the Provider and Rotator interfaces are disjoint, Rotator.Open takes inputs + state, and Provider just takes inputs.

eval/patch.go
dschaller: Yeah or just rename to something like `ApplyValuePatches`

pgavlin:
I don't love the short form intermixing `state` and inputs. I would prefer that the short form just lifts `provider` from a property to part of the function name and retains both `inputs` and `state` as properties. This allows users to write things like:

```yaml
  commonInputs:
    ...
  foo:
    fn::rotate::provider:
      inputs: ${commonInputs}
      state: ...
```

I think this is kind of a fundamental issue, FWIW. Even if we store state out-of-band, unless we reference it by path (which kills the ability to refactor via cut-and-paste), we'd need something in the provider inputs that serves as an identifier for the state.

ast/expr.go
```diff
@@ -696,6 +738,78 @@ func parseShortOpen(node *syntax.ObjectNode, name *StringExpr, args Expr) (Expr,
...
+	if !ok && state != nil {
+		diags.Extend(ExprError(stateExpr, "rotation state must be an object literal"))
+	}
+
+	return RotateSyntax(node, name, obj, provider, inputs, state), diags
+}
+
+func parseShortRotate(node *syntax.ObjectNode, name *StringExpr, args Expr) (Expr, syntax.Diagnostics) {
+	kvp := node.Index(0)
+	provider := StringSyntaxValue(name.Syntax().(*syntax.StringNode), strings.TrimPrefix(kvp.Key.Value(), "fn::rotate::"))
```
dschaller, line 787: super super nit: not sure if we want to update these variables to be rotator instead of provider now that the schema is updated

ast/expr.go
```diff
@@ -696,6 +738,78 @@ func parseShortOpen(node *syntax.ObjectNode, name *StringExpr, args Expr) (Expr,
...
+	}); i != -1 {
+		stateExpr = inputs.Entries[i].Value
+		inputs.Entries = slices.Delete(inputs.Entries, i, i+1)
+	}
+
+	state, ok := stateExpr.(*ObjectExpr)
+	if !ok && state != nil {
+		diags := syntax.Diagnostics{ExprError(stateExpr, "rotation state must be an object literal")}
+		return RotateSyntax(node, name, args, nil, nil, nil), diags
+	}
```
dschaller, line 808: Just want to check, we are ok if there is no initial state set on the rotator? (This makes sense but I vaguely recall us mentioning that users would set at least current on this when adding a new rotator)

eval/eval.go
```diff
@@ -932,6 +969,86 @@ func (e *evalContext) evaluateBuiltinOpen(x *expr, repr *openExpr) *value {
...
+			inputs.export("").Value.(map[string]esc.Value),
+			state.export("").Value.(map[string]esc.Value),
+		)
+		if err != nil {
+			e.errorf(repr.syntax(), "rotate: %s", err.Error())
+			v.unknown = true
+			return v
+		}
+
+		// todo: validate newState conforms to state schema
```
dschaller, line 1026: should we cut a ticket for this and link to it here?

eval/eval.go
```diff
@@ -932,6 +969,86 @@ func (e *evalContext) evaluateBuiltinOpen(x *expr, repr *openExpr) *value {
...
+		v.unknown = true
+		return v
+	}
+
+	// if rotating, invoke prior to open
+	if e.rotating {
+		newState, err := rotator.Rotate(
+			e.ctx,
+			inputs.export("").Value.(map[string]esc.Value),
+			state.export("").Value.(map[string]esc.Value),
```
dschaller, line 1018: Just for my own knowledge, what does `export("")` do?

eval/eval.go
nyobe: I just don't know how to do it 😢

nyobe:
> I don't love the short form intermixing `state` and inputs. I would prefer that the short form just lifts `provider` from a property to part of the function name and retains both `inputs` and `state` as properties. 

Yeah, hoisting the state out of inputs for the short form is a bit cheeky. I've removed it: https://github.com/pulumi/esc/pull/432/commits/53dc66cb807b43ffb7bc797dff6944e6959251bd

ast/expr.go
nyobe: I think it depends on the Rotator- the schema can be used to enforce it being present or optional

eval/eval.go
nyobe: https://github.com/pulumi/esc/issues/435

eval/eval.go
nyobe: This is supposed to be the environment name which I think is just used for annotating source ranges?

ast/expr.go
nyobe: 733317330fd376a0411fcb91ee758e436abf341d

ast/expr.go
```diff
@@ -434,6 +434,41 @@ func Open(provider string, inputs *ObjectExpr) *OpenExpr {
 	}
 }
 
+// RotateExpr is a type of OpenExpr that supports a rotate operation.
+type RotateExpr struct {
+	builtinNode
+
+	Rotator *StringExpr
```
pgavlin, line 441: nit: would prefer `Provider`

ast/expr.go
```diff
@@ -434,6 +434,41 @@ func Open(provider string, inputs *ObjectExpr) *OpenExpr {
...
+		Inputs:      inputs,
+		State:       state,
+	}
+}
+
+func Rotate(rotator string, inputs, state *ObjectExpr) *RotateExpr {
+	name, rotatorX := String("fn::rotate"), String(rotator)
+
+	entries := []ObjectProperty{
+		{Key: String("rotator"), Value: rotatorX},
```
pgavlin, line 459:
```suggestion
		{Key: String("provider"), Value: rotatorX},
```

ast/expr.go
```diff
@@ -696,6 +737,86 @@ func parseShortOpen(node *syntax.ObjectNode, name *StringExpr, args Expr) (Expr,
...
+	}
+
+	var rotatorExpr, inputs, stateExpr Expr
+	var diags syntax.Diagnostics
+
+	for i := 0; i < len(obj.Entries); i++ {
+		kvp := obj.Entries[i]
+		key := kvp.Key
+		switch key.GetValue() {
+		case "rotator":
```
pgavlin, line 754: nit: I'd prefer "provider" in place of "rotator". the only disadvantage I see to using provider instead of rotator is that it could be read as implying that all providers are rotatable, which is not the case. the advantage is that it retains some additional symmetry with `fn::open`.

```suggestion
		case "provider":
```

pgavlin: (in any case we can always change this based on feedback)

eval/eval.go
pgavlin: yep that's right

eval/eval.go
```diff
@@ -932,6 +969,87 @@ func (e *evalContext) evaluateBuiltinOpen(x *expr, repr *openExpr) *value {
...
+			e.errorf(repr.syntax(), "rotate: %s", err.Error())
+			v.unknown = true
+			return v
+		}
+
+		// todo: validate newState conforms to state schema
+
+		e.patchOutputs = append(e.patchOutputs, &Patch{
+			// rotation output is written back to the fn's `state` input
+			DocPath:     x.path + "." + repr.node.Name().GetValue() + ".state",
```
pgavlin, line 1031: probably safer to do something like this, which will apply any necessary escaping to `repr.node.Name().GetValue()`:

```suggestion
			DocPath:     util.JoinKey(x.path, repr.node.Name().GetValue()) + ".state",
```

provider.go
```diff
@@ -29,3 +29,16 @@ type Provider interface {
 	// Open retrieves the provider's secrets.
 	Open(ctx context.Context, inputs map[string]Value, executionContext EnvExecContext) (Value, error)
 }
+
+// A Rotator enables environments to rotate a secret.
+// Rotators persist internal state by writing back to the environment.
```
pgavlin, line 34: should rephrase this doc comment along the lines of "it is the responsibility of the caller to appropriately persist rotation state (e.g. by writing it back to the environment definition).

pgavlin (approved):
one nit on naming, but LGTM otherwise

dschaller (approved):

PR #433, opened 2025-01-15
Add rotate command

seanyeh:
## Testing

`go run ./cmd/esc env rotate temp2/abc/rotate`
Rotated and opened the env. Also tested with various `-f` options

Does not accept `@version`
```
❯ go run ./cmd/esc env rotate temp2/abc/yo@asdf
Error: the rotate command does not accept environments at specific versions
exit status 1
```

### DIFF
--- a/cmd/esc/cli/cli_test.go
+++ b/cmd/esc/cli/cli_test.go
@@ -677,6 +677,16 @@ func (c *testPulumiClient) OpenEnvironment(
 	return c.openEnvironment(ctx, orgName, envName, env.yaml)
 }
 
+func (c *testPulumiClient) RotateEnvironment(
+	ctx context.Context,
+	orgName string,
+	projectName string,
+	envName string,
+	duration time.Duration,
+) (string, []client.EnvironmentDiagnostic, error) {
+	return c.OpenEnvironment(ctx, orgName, projectName, envName, "", duration)
+}
+
 func (c *testPulumiClient) CheckYAMLEnvironment(
 	ctx context.Context,
 	orgName string,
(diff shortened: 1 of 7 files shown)


nyobe (approved):

cmd/esc/cli/cli_test.go
```diff
@@ -677,6 +677,16 @@ func (c *testPulumiClient) OpenEnvironment(
...
 }
 
+func (c *testPulumiClient) RotateEnvironment(
+	ctx context.Context,
+	orgName string,
+	projectName string,
+	envName string,
+	duration time.Duration,
+) (string, []client.EnvironmentDiagnostic, error) {
+	return c.OpenEnvironment(ctx, orgName, projectName, envName, "", duration)
```
IaroslavTitov, line 687: Hmm, not sure I understand why RotateEnvironment calls open? Is this a typo?

cmd/esc/cli/client/client.go
```diff
@@ -616,6 +630,37 @@ func (pc *client) OpenEnvironment(
...
+	duration time.Duration,
+) (string, []EnvironmentDiagnostic, error) {
+	path := fmt.Sprintf("/api/esc/environments/%v/%v/%v/rotate", orgName, projectName, envName)
+
+	queryObj := struct {
+		Duration string `url:"duration"`
+	}{
+		Duration: duration.String(),
+	}
+	var resp struct {
```
IaroslavTitov, line 647: Nit: could maybe move these queryObj and resp structs out to reuse between functions, not sure how many times these are used

cmd/esc/cli/cli_test.go
seanyeh: this is for the test client for tests. I don't know exactly how it should do a fake "rotate" so I'm assuming a no-op and calling open (since rotating calls open)

IaroslavTitov (approved):

PR #436, opened 2025-01-25
fix optional rotation state

nyobe:
rotation state was intended to be optional, so that it could be left off a new `fn::rotate` if not bootstrapping initial values, but this doesn't actually work currently.

leaving state off [does parse](https://github.com/pulumi/esc/blob/main/ast/expr.go#L776-L777), but it ends up evaluating to an unknown value (since it's a missingExpr) which causes the evaluator to [bail out](https://github.com/pulumi/esc/blob/d038e93ef5c8b6a477bae2cc48d30c23f5eaa6e8/eval/eval.go#L1008).

so this tries to fix that by treating a missing state key as null, and relaxing the rotator schema to always allow state to be null.

~~but now that I've done this, I am having second thoughts about the whole exercise... maybe the `state` key should just be unconditionally required like the `inputs` key is, and the provider's schema can decide whether it wants to allow empty state objects or not?  then a user will always need to provide at least `state: {}`, but maybe that's fine, since state is a fundamental part of this type of provider.~~


seanyeh (approved):
tested this with some env defs, confirmed state gets created when not provided, and schema validation is fixed (requiring `inputs`)

PR #437, opened 2025-01-27
support rotating by path

nyobe:
Re-add support for specifying a subset of paths to rotate to #432.

Closes https://github.com/pulumi/pulumi-service/issues/24986


eval/eval.go
```diff
@@ -115,8 +115,13 @@ func RotateEnvironment(
 	providers ProviderLoader,
 	environments EnvironmentLoader,
 	execContext *esc.ExecContext,
+	paths []string,
```
pgavlin, line 118: Might want this to be a list of parsed paths (e.g. `[]resource.PropertyPath` or something). At the very least we should parse these for validation and normalization prior to adding them to the et.

eval/testdata/eval/rotate/overrides.json
```diff
@@ -1 +1 @@
-{"rotate": true}
+{"rotate": true, "rotatePaths": ["full", "short"]}
```
pgavlin, line 1: would love to see some more complex paths here--e.g. with `quoted["property \"accessors\""]`, indices into lists, and normal dotted property access

nyobe: I've added more complex paths, and combined with parsing property paths it works well, thank you.

However, I tried to include one that is embedded within another function, but it seems like the expr paths in the evaluator are getting reset when parsing function arguments.  _should_ that be expected to work? (eg the expr's path when evaluating the `fn::open::test.some-input` is just `some-input`)

pgavlin: Hmmm interesting. That's worth some more digging.

So there are two sorts of paths at play here: value paths and document paths. The former are the paths used in `${symbols.and.interpolations}` and are resolved with respect to the schema of the fully-evaluated environment. The latter are resolved with respect to the YAML document. We want these paths to be document paths. I'm not sure what sort the paths on the expr nodes are (or what they're intended to be).

nyobe: okay, I'm not sure, but I think that maybe `expr.path` is might be intended to be _value path_, based on the comment that it's [for reporting cyclic reference errors](https://github.com/pulumi/esc/blob/6198f06dce04275d8dfb06b0b5fa433088989d1a/eval/expr.go#L36).  The path is built up by `declare`, which [does reset the path when recursing into fn arguments](https://github.com/pulumi/esc/blob/a82231749da7cad36f2ff912816eaab922378e42/eval/eval.go#L343).. But I think this might not be quite right even for value paths, since they should still be qualified by the parent path, but just eliding the function name, right?

It looks like I might be able to recover the _document path_ from the ast node, using `x.repr.syntax().Syntax().Syntax().Path()` (which resolves to expr -> exprRepr -> ast.Expr -> syntax.Node -> syntax.Syntax (encoding.YAMLSyntax))`

seanyeh (approved):

eval/testdata/eval/rotate-paths/overrides.json
```diff
@@ -0,0 +1,8 @@
+{
+  "rotate": true,
+  "rotatePaths": [
+    "examples[\"subscript-path\"][0]",
+    "examples.deeply.nested[0].quoted \"property\".path",
```
pgavlin, line 5: 😮 I'm surprised this works to be honest. would have expected this to require `examples.deeply.nested[0]["quoted \"property\"].path`

pgavlin (approved):
One note about a surprising test case, LGTM otherwise

eval/testdata/eval/rotate-paths/overrides.json
nyobe: hmm, yeah. it looks like the [parser is a bit lenient here](https://github.com/pulumi/pulumi/blob/master/sdk/go/common/resource/properties_path.go#L130-L137) and consumes everything up to the next `.` even though this isn't strictly legal..  I'll replace it with an actually valid path 😅

pgavlin: Might be worth considering a break in the future to fix the parser. This shouldn't be allowed.

PR #438, opened 2025-01-28
teach evaluator to conditionally retry provider invocation

nyobe:
Adds a `RetryableError` error that providers can return to opt into being retried by the evaluator when they encounter a temporary error (such as a timeout).

Closes https://github.com/pulumi/pulumi-service/issues/25444

### DIFF
--- a/provider.go
+++ b/provider.go
@@ -42,3 +42,17 @@ type Rotator interface {
 	// Rotate rotates the provider's secret, and returns the rotator's new state to be persisted.
 	Rotate(ctx context.Context, inputs, state map[string]Value, executionContext EnvExecContext) (Value, error)
 }
+
+// RetryableError indicates a temporary error was encountered by a provider.
+// Providers can wrap an error with this type  to indicate to the evaluator that it may retry calling the provider.
+type RetryableError struct {
+	Err error
+}
+
+func (e RetryableError) Error() string {
+	return e.Err.Error()
+}
+
+func (e RetryableError) Unwrap() error {
+	return e.Err
+}
--- a/provider.go
+++ b/provider.go
@@ -16,6 +16,7 @@ package esc
 
 import (
 	"context"
+	"errors"
 
 	"github.com/pulumi/esc/schema"
 )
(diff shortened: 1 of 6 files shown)


provider.go
```diff
@@ -42,3 +42,17 @@ type Rotator interface {
 	// Rotate rotates the provider's secret, and returns the rotator's new state to be persisted.
 	Rotate(ctx context.Context, inputs, state map[string]Value, executionContext EnvExecContext) (Value, error)
 }
+
+// RetryableError indicates a temporary error was encountered by a provider.
+// Providers can wrap an error with this type  to indicate to the evaluator that it may retry calling the provider.
+type RetryableError struct {
+	Err error
```
pgavlin, line 49: nit: since we have `Unwrap`, I'd make this field unexported

pgavlin: also--let's add a factory for these

```
func Retryable(err error) RetryableError { return RetryableError{err} }
```

pgavlin (commented):
so at a high level I'm wondering if this is kind of a footgun.

if a rotator gets halfway through a rotation, it'll need to have some way to understand the state it is in when it is next retried.

I don't think that we have a solution for that at the moment.

I wonder if instead of handling this in the evaluator we should just add a helper function that provider authors can use for retries.

nyobe:
> if a rotator gets halfway through a rotation, it'll need to have some way to understand the state it is in when it is next retried.

When providers have access to managing creds they'll usually have the authority to overwrite whatever they find, so keeping intermediate state isn't as necessary like it would be if we tried to use the rotated credential to manage itself.

> I wonder if instead of handling this in the evaluator we should just add a helper function that provider authors can use for retries.

We sort of do have a helper available with the `retry.Until` utility.  Also, many sdks (like aws's) do their own builtin retries.  Since the provider itself has the most context about the errors it encounters, maybe we should just let the provider itself make the decision on how it wants to retry itself, rather than trying to put that in the evaluator?  (So just close this for now?)

provider.go
nyobe: 33eda08026f289e767804f11c51126eea21682b0

pgavlin:
> Since the provider itself has the most context about the errors it encounters, maybe we should just let the provider itself make the decision on how it wants to retry itself, rather than trying to put that in the evaluator? (So just close this for now?)

Yeah, this is what I'm thinking tbh. That allows us to sidestep questions of state, etc.

PR #439, opened 2025-01-28
add new import type that is only resolved during rotation

nyobe:
[Design doc](https://docs.google.com/document/d/1qWOngCTaiSutunRq0Kx_Axnc-8oI3dOFUDa2yA1_c9Q/edit?tab=t.0#heading=h.i4diafbtsiun)

Add an additional `rotateOnly` metadata for imports, (similar to merge.)  Imports with `rotateOnly` set to true will only be evaluated during rotation, not during open, and are otherwise treated as unknown.

Additionally, `fn::rotate` now tolerates unknown inputs so that `open` will succeed when using values from such imports.

Combining these allows rotation credentials to be imported from a restricted environment, while enabling the rotated environment to be consumed by lower-privileged users.
```
imports:
- foo: {rotateOnly: true}

values:
  example:
    fn::rotate::provider:
      inputs: ${some_creds_from_foo}
      state: {}
```

Resolves https://github.com/pulumi/pulumi-service/issues/25240


eval/eval.go
```diff
@@ -482,6 +486,9 @@ func (e *evalContext) evaluateImport(myImports map[string]*value, decl *ast.Impo
 			return
 		}
 		val = imported.value
+	} else if rotateOnly && !e.rotating {
+		// this import should only be evaluated during rotation, and we're not rotating, so resolve it as unknown
+		val = &value{def: newMissingExpr("", nil), schema: schema.Always(), unknown: true}
```
nyobe, line 491: I know we talked about adding a new type of missingExpr for this, but I haven't quite wrapped my head around how it would work.  

I think we need to keep track of the imported environment name somewhere? I'd like for a rotation provider to be able to return an error saying it tried to use an unknown value / you may not have permission to access <env>, but I'm not sure how to recover that information from esc.Value.

pgavlin: Ah, I'd put that in the evaluator itself--if the unknown value is accessed when not rotating, the evaluator should error. This could be done via a patch to `evaluateExprAccess` to handle the distinguished expression type in its type switch.

pgavlin: Hm, though I suppose that wouldn't work, since it's not clear whether or not the value is actually _used_ by the provider. Ick.

pgavlin: I almost wonder if we want rotators to have two input schemas? one that's user-facing and one that's internal? that would at least allow the evaluator to handle this instead of the rotator.

nyobe: yeah, the evaluator doesn't have the context to know if the value is _used_ or not. 

adding a second input schema can help give us this context, we could have one input schema that is validated during open and another that is validated during rotate.

this almost works, except that it seems like unknown values are not being treated as missing for required properties, so it's still validating successfully 😕  https://github.com/pulumi/esc/pull/439/commits/65c34a527511f78657af8421025d361ddf0e43f8

nyobe:
closing in favor of #443 + #444

PR #443, opened 2025-02-12
inline environment references

nyobe:
The `environments` top level key is now reserved for inline environment references. This allows you to implicitly import and reference a value in another environment by accessing `${environments.$project.$env...}`. This desugars into something like `${imports["$project/$env"]...}` and behaves as if the environment was imported with merge:false.

This is followed up by #444

### DIFF
--- a/eval/eval.go
+++ b/eval/eval.go
@@ -390,7 +390,7 @@ func declare[Expr exprNode](e *evalContext, path string, x Expr, base *value) *e
 
 func (e *evalContext) isReserveTopLevelKey(k string) bool {
 	switch k {
-	case "imports", "context":
+	case "imports", "context", "environments":
 		return true
 	default:
 		return false
@@ -699,6 +699,11 @@ func (e *evalContext) evaluateExprAccess(x *expr, accessors []*propertyAccessor)
 		return e.evaluateValueAccess(x.repr.syntax(), e.myContext, accessors[1:])
 	}
 
+	// Check for inline reference
+	if ok && k == "environments" {
+		return e.evaluateEnvironmentReferenceAccess(x, accessors)
+	}
+
 	for len(accessors) > 0 {
 		accessor := accessors[0]
 		if receiver == nil {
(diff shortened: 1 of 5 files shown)


eval/eval.go
```diff
@@ -751,6 +756,61 @@ func (e *evalContext) evaluateExprAccess(x *expr, accessors []*propertyAccessor)
 	return e.evaluateExpr(receiver)
 }
 
+// evaluateEnvironmentReferenceAccess performs an inline import of an environment.
+// The sugared accessor is of the form ["environments", $project, $env, ...], which is desugared into the import name
```
pgavlin, line 760: not sure that I'd consider this sugar--there's no equivalent expression a user could write (e.g. `${environments.foo.bar}` is not the same as `${environments["foo/bar"]}`).

eval/eval.go
```diff
@@ -751,6 +756,61 @@ func (e *evalContext) evaluateExprAccess(x *expr, accessors []*propertyAccessor)
...
+	qualifiedName := fmt.Sprintf("%s/%s", projName, envName)
+
+	// create an isolated value to hold the inline import map
+	inlineImports := map[string]*value{}
+	e.evaluateImport(inlineImports, &ast.ImportDecl{
+		Meta: &ast.ImportMetaDecl{
+			Merge: ast.Boolean(false),
+		},
+		Environment: ast.String(qualifiedName),
+	})
```
pgavlin, line 781: rather than synthesizing a decl, I think I'd pull https://github.com/pulumi/esc/blob/main/eval/eval.go#L488-L516 out into its own method (maybe `evaluateBareImport` or something?), then call that here. Slightly tricky but I think we'll end up with a better experience (e.g. as-is I think error reporting is going to be off, since we're not handing out real syntax w/ location info).

the extracted method could also return a `bool` to indicate whether or not evaluation succeeded and we could avoid the indirect check below.

pgavlin (commented):
this is so much slicker that our other options. one comment about a refactoring but the core looks good to me.

eval/eval.go
nyobe: I just meant sugar insofar as creating the actual import name

eval/eval.go
nyobe: good call, the errors look much better now: https://github.com/pulumi/esc/pull/443/commits/3d5619df4def71ccc01c67a9aee658fd2bff66ba#diff-e1baf80911fb297e49a025b34259d0458f79c636efcd26a681ac88d36b619933L12

pgavlin (approved):

PR #445, opened 2025-02-14
schema: rotateOnly flag is now represented as a list of property names

nyobe:
The rotateOnly flag is now represented by a list of property names on the parent object schema, similar to `required`.

During schema compilation, the flag is internally pushed down onto the properties, and is accessible via their IsRotateOnly() method.


pgavlin (approved):

pgavlin:
Please update the title and add a description of the change.

PR #448, opened 2025-02-18
WIP - Surface rotation results in eval

seanyeh:


eval/eval.go
```diff
@@ -193,12 +195,27 @@ type evalContext struct {
...
+	Status RotationStatus
+	Error  string
+}
+
+type RotationStatus string
+
+const (
+	RotationSucceeded    RotationStatus = "succeeded"
+	RotationFailed       RotationStatus = "failed"
+	RotationNotEvaluated RotationStatus = "not_evaluated"
```
pgavlin, line 216: tiny nit:

```suggestion
	RotationNotEvaluated RotationStatus = "not-evaluated"
```

eval/eval.go
```diff
@@ -193,12 +195,27 @@ type evalContext struct {
...
-	rotateDocPaths map[string]bool // the subset of document paths to invoke rotation for when rotating. if empty, all rotators will be invoked.
-	patchOutputs   []*Patch        // updated rotation state generated during evaluation, to be written back to the environment definition
+	rotateDocPaths  map[string]bool   // the subset of document paths to invoke rotation for when rotating. if empty, all rotators will be invoked.
+	patchOutputs    []*Patch          // updated rotation state generated during evaluation, to be written back to the environment definition
+	rotationResults []*RotationResult // list of results of rotations performed
 
 	diags syntax.Diagnostics // diagnostics generated during evaluation
 }
 
+type RotationResult struct {
```
pgavlin, line 205: Please add a doc comment to this type + its fields

eval/eval.go
```diff
@@ -193,12 +195,27 @@ type evalContext struct {
...
+	Path   string
+	Status RotationStatus
+	Error  string
+}
+
+type RotationStatus string
```
pgavlin, line 211: Same request re: doc comment

eval/eval.go
```diff
@@ -193,12 +195,27 @@ type evalContext struct {
...
+	patchOutputs    []*Patch          // updated rotation state generated during evaluation, to be written back to the environment definition
+	rotationResults []*RotationResult // list of results of rotations performed
 
 	diags syntax.Diagnostics // diagnostics generated during evaluation
 }
 
+type RotationResult struct {
+	Path   string
+	Status RotationStatus
+	Error  string
```
pgavlin, line 208: Instead of an error string, let's return a slice of diagnostics. That should be a bit more future-proof.

eval/eval.go
```diff
@@ -194,11 +194,32 @@ type evalContext struct {
...
+type RotationResult struct {
+	Rotations    []*Rotation // rotations performed
+	PatchOutputs []*Patch    // updated rotation state generated during evaluation, to be written back to the environment definition
+}
+
+// A Rotation stores secret rotation information and diagnostics
+type Rotation struct {
+	Path   string
+	Status RotationStatus
+	Diags  syntax.Diagnostics
```
nyobe, line 212: I think it might make sense to have a `Patch` field here, as part of `Rotation`, instead of a separate array of patches in `RotationResult`.  Then for each result you'd have either diags or a patch.

seanyeh: that's a good idea

eval/eval.go
```diff
@@ -1120,18 +1128,31 @@ func (e *evalContext) evaluateBuiltinRotate(x *expr, repr *rotateExpr) *value {
 			e.execContext,
 		)
 		if err != nil {
+			diag := ast.ExprError(repr.syntax(), err.Error())
+			rotation := Rotation{
+				Path:   docPath,
+				Status: RotationFailed,
+				Diags:  []*syntax.Diagnostic{diag},
+			}
+			e.rotationResult.Rotations = append(e.rotationResult.Rotations, &rotation)
```
nyobe, line 1137:
```suggestion
			 e.rotationResult.Rotations = append(e.rotationResult.Rotations, &Rotation{
				Path:   docPath,
				Status: RotationFailed,
				Diags:  []*syntax.Diagnostic{diag},
			})
```

nit, just so it's the same as above ^
same for the following one

eval/rotation.go
```diff
@@ -0,0 +1,49 @@
...
+const (
+	RotationSucceeded    RotationStatus = "succeeded"
+	RotationFailed       RotationStatus = "failed"
+	RotationNotEvaluated RotationStatus = "not-evaluated"
+)
+
+// A RotationResult stores the result of secret rotations
+type RotationResult struct {
+	Rotations []*Rotation // rotations performed
+}
```
nyobe, line 30: this could just be `type RotationResults []*Rotation`

eval/eval_test.go
```diff
@@ -404,9 +404,11 @@ func TestEval(t *testing.T) {
 				var rotated *esc.Environment
 				var patches []*Patch
 				var rotateDiags syntax.Diagnostics
+				var rotationResult *RotationResult
 				if doRotate {
-					rotated, patches, rotateDiags = RotateEnvironment(context.Background(), environmentName, env, rot128{}, testProviders{},
+					rotated, rotationResult, rotateDiags = RotateEnvironment(context.Background(), environmentName, env, rot128{}, testProviders{},
 						&testEnvironments{basePath}, execContext, rotatePaths)
+					patches = rotationResult.Patches()
```
nyobe, line 411: we might consider updating the snapshots json to contain the full rotationResult, rather than just patches. I think we should do that in a followup, because it's good to see that the tests still pass as is with this refactor.

nyobe (approved):
nice!

eval/eval.go
```diff
@@ -117,7 +117,7 @@ func RotateEnvironment(
 	environments EnvironmentLoader,
 	execContext *esc.ExecContext,
 	paths []resource.PropertyPath,
-) (*esc.Environment, []*Patch, syntax.Diagnostics) {
```
nyobe, line 120: now that RotationResult is a slice, we shouldn't use a pointer to it

eval/eval.go
```diff
@@ -165,7 +165,7 @@ func evalEnvironment(
 		Properties:       v.export(name).Value.(map[string]esc.Value),
 		Schema:           s,
 		ExecutionContext: executionContext,
-	}, ec.patchOutputs, diags
+	}, &ec.rotationResult, diags
```
nyobe, line 168: now that RotationResult is a slice, we shouldn't wrap it with an additional pointer, just return the nil slice itself

PR #451, opened 2025-02-19
eval: containsObservableUnknowns is incorrect for non-rotateOnly inputs

nyobe:
The intent of `containsObservableUnknowns` is to allow `evaluateBuiltinRotate` tolerate rotateOnly values being unknown when not rotating.  RotateOnly indicates that `open` will not read those values, so it's okay for them to be unknown.  

During rotation, `rotate` will read all inputs, so all inputs must be known at that time.

Unfortunately the equality isn't correct, because it's returning false for unknown, non-rotateOnly values, which allows for an unknown value to be unexpectedly passed into the provider's Rotate method.

Fixes https://github.com/pulumi/pulumi-service/issues/26101

### DIFF
--- a/eval/value_test.go
+++ b/eval/value_test.go
@@ -75,3 +75,42 @@ func TestValueToString(t *testing.T) {
 		})
 	}
 }
+
+func TestContainsObservableUnknowns(t *testing.T) {
+	fake := newMissingExpr("", nil)
+	cases := map[string]struct {
+		value    *value
+		expected bool
+	}{
+		"unknown is rotateOnly, so it's not observable": {
+			value: &value{def: fake, repr: map[string]*value{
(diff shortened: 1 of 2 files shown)


eval/value_test.go
```diff
@@ -75,3 +75,42 @@ func TestValueToString(t *testing.T) {
...
+			value: &value{def: fake, repr: map[string]*value{
+				"rotateOnly": {def: fake, rotateOnly: true, unknown: false, repr: "foo"},
+				"plain":      {def: fake, rotateOnly: false, unknown: false, repr: "bar"},
+			}},
+			expected: false,
+		},
+	}
+	for name, c := range cases {
+		t.Run(name, func(t *testing.T) {
+			// when rotating, containsObservableUnknowns === containsUnknowns
```
pgavlin, line 109: excellent to capture this condition 👍

PR #463, opened 2025-02-26
Fix setting values starting with `[` or `]`

seanyeh:
Fixes error when trying to run `esc env set <env> <path> "[a"`

### DIFF
--- a/cmd/esc/cli/env_set.go
+++ b/cmd/esc/cli/env_set.go
@@ -59,7 +59,15 @@ func newEnvSetCmd(env *envCommand) *cobra.Command {
 
 			var yamlValue yaml.Node
 			if err := yaml.Unmarshal([]byte(args[1]), &yamlValue); err != nil {
-				return fmt.Errorf("invalid value: %w", err)
+				// Default to a string value
+				yamlValue = yaml.Node{
+					Content: []*yaml.Node{
+						{
+							Kind:  yaml.ScalarNode,
+							Value: args[1],
+						},
+					},
+				}
 			}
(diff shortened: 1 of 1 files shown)


cmd/esc/cli/env_set.go
```diff
@@ -59,7 +59,15 @@ func newEnvSetCmd(env *envCommand) *cobra.Command {
 
 			var yamlValue yaml.Node
 			if err := yaml.Unmarshal([]byte(args[1]), &yamlValue); err != nil {
-				return fmt.Errorf("invalid value: %w", err)
+				// Default to a string value
+				yamlValue = yaml.Node{
```
seanyeh, line 63: @pgavlin are there any cases where we **don't** want to default to the string value if the given value can't be parsed in YAML?

pgavlin: My gut is to say "all cases, as this defaulting makes the behavior harder to understand". I can imagine adding a flag (e.g. `--string`) that forces the ESC CLI to treat its input as a string and not attempt to parse it as YAML at all.

pgavlin: something like this: https://github.com/pulumi/esc/compare/pgavlin/env-set-string

seanyeh:
closing in favor of https://github.com/pulumi/esc/pull/467

cmd/esc/cli/env_set.go
seanyeh: @pgavlin i made a [PR](https://github.com/pulumi/esc/pull/467) out of your branch

PR #467, opened 2025-02-27
cli: add a --string flag to env set

seanyeh:
Add a `--string` flag to `env set` to treat given value as a string

### DIFF
--- a/cmd/esc/cli/env_set.go
+++ b/cmd/esc/cli/env_set.go
@@ -4,6 +4,7 @@ package cli
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"regexp"
 	"strconv"
@@ -20,6 +21,7 @@ import (
 func newEnvSetCmd(env *envCommand) *cobra.Command {
 	var secret bool
 	var plaintext bool
+	var rawString bool
 
 	cmd := &cobra.Command{
 		Use:   "set [<org-name>/][<project-name>/]<environment-name> <path> <value>",
@@ -57,8 +59,17 @@ func newEnvSetCmd(env *envCommand) *cobra.Command {
 				return fmt.Errorf("path must contain at least one element")
 			}
 
+			input := args[1]
+			if rawString {
+				js, err := json.Marshal(input)
+				if err != nil {
+					return fmt.Errorf("internal error: %w", err)
+				}
+				input = string(js)
+			}
+
 			var yamlValue yaml.Node
-			if err := yaml.Unmarshal([]byte(args[1]), &yamlValue); err != nil {
+			if err := yaml.Unmarshal([]byte(input), &yamlValue); err != nil {
 				return fmt.Errorf("invalid value: %w", err)
 			}
 			if len(yamlValue.Content) == 0 {
@@ -147,6 +158,9 @@ func newEnvSetCmd(env *envCommand) *cobra.Command {
 	cmd.Flags().BoolVar(
 		&plaintext, "plaintext", false,
 		"true to leave the value in plaintext")
+	cmd.Flags().BoolVar(
+		&rawString, "string", false,
+		"true to treat the value as a string rather than attempting to parse it as YAML")
 
 	return cmd
 }
--- a/cmd/esc/cli/env_set.go
+++ b/cmd/esc/cli/env_set.go
@@ -4,10 +4,10 @@ package cli
 
 import (
 	"context"
-	"encoding/json"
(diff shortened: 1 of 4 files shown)


cmd/esc/cli/env_set.go
```diff
@@ -57,8 +59,17 @@ func newEnvSetCmd(env *envCommand) *cobra.Command {
 				return fmt.Errorf("path must contain at least one element")
 			}
 
+			input := args[1]
+			if rawString {
+				js, err := json.Marshal(input)
```
nyobe, line 64: yaml.Node has a [SetString](https://github.com/pulumi/esc/pull/464/commits/db51cc65f82f9d058e7c8d13c701c39562b854f7#diff-b68423f1a4998ad681d998d214ba2230c8a74b91c2ec03f4ec9e7fc4a5c71209R90-R91) helper that you can use, rather than this marshal/unmarshal round trip.

this also sets the block style automatically.  a [slight change](https://github.com/pulumi/esc/pull/464/commits/d3c8be4d5f08075e9cfe243ee3a584731cff30e2#diff-8e05dbaf49b50bc2082e21df6ea853d0051dceaa9de1cfbbfb9b16915d99f3e4R139) lets us preserve that so we use the nicer block string syntax when there are newlines, rather than escaping them like the json marshaling does

pgavlin: I worry that we need to be a little bit more sophisticated than `SetString`. That helper doesn't handle non-printable characters, for example (literal style scalars don't offer any escape sequences): https://github.com/go-yaml/yaml/blob/v3.0.1/yaml.go#L471-L483

We could use that method and then override the quoting style for strings with non-printable characters:

```
node.SetString(input)
if mustEscape := strings.ContainsFunc(input, func(r rune) bool { return !strconv.IsPrint(r) }); mustEscape {
    node.Style = yaml.DoubleQuotedStyle
}
```

cmd/esc/cli/testdata/env-set-string.yaml
```diff
@@ -0,0 +1,10 @@
+run: |
+  esc env init default/test
+  esc env set default/test path "[a" --string
```
pgavlin, line 3: This test should also run an `esc env get` to ensure that the string's content is as expected

cmd/esc/cli/env_set.go
pgavlin: Yeah that's not an adequate check. That'll catch invalid UTF-8 but not valid UTF-8 with non-printable chars.

cmd/esc/cli/env_set.go
```diff
@@ -57,17 +59,25 @@ func newEnvSetCmd(env *envCommand) *cobra.Command {
...
-			if len(yamlValue.Content) == 0 {
-				// This can happen when the value is empty (e.g. when "" is present on the command line). Treat this
-				// as the empty string.
-				err = yaml.Unmarshal([]byte(`""`), &yamlValue)
-				contract.IgnoreError(err)
+			if rawString {
+				yamlValue.SetString(input)
+				if mustEscape := strings.ContainsFunc(input, func(r rune) bool { return !strconv.IsPrint(r) }); mustEscape {
+					yamlValue.Style = yaml.DoubleQuotedStyle
+				}
```
pgavlin, line 68: Out of paranoia--have we verified that this works for non-printable characters? I _think_ it will--it should--but maybe we need to manually escape these characters as well as setting the literal style?

pgavlin (commented):
One paranoid question, but LGTM otherwise

cmd/esc/cli/testdata/env-set-string.yaml
```diff
@@ -0,0 +1,14 @@
...
+  esc env get default/test path --value string
```
pgavlin, line 4: Oh--nit that we should grab the whole env so that we can see the exact way this gets serialized

cmd/esc/cli/env_set.go
seanyeh: @pgavlin Yah, this seems to work.
For example:
`go run ./cmd/esc env set delme/abc/audit a "$(echo "Hello\x05World")" --string`
becomes
`a: "Hello\x05World"` (this is correct, right?)

pgavlin: Rad. Yeah that's good. Could we add that as a test case?

pgavlin (approved):

PR #471, opened 2025-03-05
Use cobra args functions instead of cmdutil (take 2)

iwahbe:
Based off of #470.

This PR migrates ESC off of `cmdutil.RunFunc` (now deprecated).


tehsis:
This PR was actually finished on #470

PR #477, opened 2025-03-10
Ensuring env var is always taken over account backend url

IaroslavTitov:
### Summary
- Fixes https://github.com/pulumi/esc/issues/473
- Makes ESC CLI always prefer env var `PULUMI_BACKEND_URL` over backend url saved into the account
- If backendURL is overridden, then login will run and look for `PULUMI_ACCESS_TOKEN` env var, falling back to accounts if it's not specified

### Testing
- Manually tested, the scenario in ticket now works
- Manually tested case when invalid backend URL is set via env var (or if user never logged in with that backend), user will see `Error: no credentials. Please run `esc login` to log in.`

### DIFF
--- a/cmd/esc/cli/login.go
+++ b/cmd/esc/cli/login.go
@@ -153,9 +153,18 @@ func (esc *escCommand) getCachedClient(ctx context.Context) error {
 	if err != nil {
 		return fmt.Errorf("could not determine current cloud: %w", err)
 	}
-	if account == nil {
-		backendURL := esc.workspace.GetCurrentCloudURL(nil)
 
+	backendURL := esc.workspace.GetCurrentCloudURL(account)
+
+	// If backendURL is overridden via env var, try to get a matching existing account
+	if backendURL != account.BackendURL {
+		account, err = esc.workspace.GetAccount(backendURL)
+		if err != nil {
+			return fmt.Errorf("could not determine current cloud: %w", err)
+		}
+	}
+
+	if account == nil {
 		nAccount, err := esc.login.Login(
 			ctx,
 			backendURL,
@@ -178,11 +187,11 @@ func (esc *escCommand) getCachedClient(ctx context.Context) error {
 
 		account = &workspace.Account{
 			Account:    *nAccount,
-			BackendURL: backendURL,
 			DefaultOrg: defaultOrg,
 		}
 	}
 
+	account.BackendURL = backendURL
 	if err := esc.checkBackendURL(account.BackendURL); err != nil {
 		return err
 	}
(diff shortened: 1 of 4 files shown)


cmd/esc/cli/login.go
```diff
@@ -153,9 +153,18 @@ func (esc *escCommand) getCachedClient(ctx context.Context) error {
...
-	if account == nil {
-		backendURL := esc.workspace.GetCurrentCloudURL(nil)
 
+	backendURL := esc.workspace.GetCurrentCloudURL(account)
+
+	// If backendURL is overridden via env var, try to get a matching existing account
+	if backendURL != account.BackendURL {
+		account, err = esc.workspace.GetAccount(backendURL)
+		if err != nil {
+			return fmt.Errorf("could not determine current cloud: %w", err)
```
ringods, line 163: This is not correct. If both `PULUMI_BACKEND_URL` and `PULUMI_ACCESS_TOKEN` are defined, and this is the first time that we use this backend, Pulumi should and will auto login using these environment variables.

ringods (changes requested):

tobiashenkel:
> @IaroslavTitov I don't agree with your statement here:
> 
> > Instead of setting a variable, user should just run `pulumi login --cloud-url ...` to switch accounts
> 
> I even have a complete opposite standpoint. If someone is working with multiple accounts, using different values for a specific environment variable allows for easy switching, definitely if you do this using [direnv](https://direnv.net/). This all becomes automatic by going into a specific directory.
> 
> I would also favor to remain consistent with the core Pulumi CLI, where I requested this via [pulumi/pulumi#13919](https://github.com/pulumi/pulumi/issues/13919) and this was integrated in Pulumi v3.130.0.

Our use case here (besides consistency with the pulumi cli): Suppose you have two shell sessions open on different stacks using different backends. Using the env var will make running the two possible in a seamless way (e.g. by using direnv) where a `pulumi login ...` call writes into a global file which makes it impossible to select the backend per session (unless you have completely different PULUMI_HOME dirs with their own config files).

cmd/esc/cli/login.go
IaroslavTitov: > I would also favor to remain consistent with the core Pulumi CLI

Sounds good, I didn't realize there's a precedent with Pulumi CLI, ignore the comment 😅 

> This is not correct IMO

I was trying to prevent user footgun, when one sets `PULUMI_BACKEND_URL`, but not the token. If that's undesirable/confusing behavior, I'll just remove this block

komalali:
Can we add tests for this?

IaroslavTitov:
> Can we add tests for this?

Done! Please take a look, recreated the scenario from the issue as a test

komalali (approved):

PR #483, opened 2025-03-11
eval, cli: better support for binary data

### LINKED ISSUE #480: Binary data corrupted during marshaling
ESC currently handles binary data by coercing binary values to a (potentially-invalid) strings. This works okay internally, since Go does not validate the wellformedness of strings when coercing from binary data. However, the standard Go JSON encoder ensures that all strings are [coerced to valid UTF-8 during marshaling](https://pkg.go.dev/encoding/json#Marshal), which replaces any invalid bytes with the replacement character. This corrupts the data.

For example, the following environment should write the bytes `dcc5 43bc 1b4f f4fb 8263 d630 a199 2688 166b c084 5b71 1240 24c3 a944 d937 e2fc` to a temporary file:

```yaml
values:
  files:
    MY_FILE:
      fn::fromBase64: 3MVDvBtP9PuCY9YwoZkmiBZrwIRbcRJAJMOpRNk34vw=
```

Instead, it writes the bytes `efbf bdef bfbd 43ef bfbd 1b4f efbf bdef bfbd efbf bd63 efbf bd30 efbf bdef bfbd 26ef bfbd 166b efbf bdef bfbd 5b71 1240 24c3 a944 efbf bd37 efbf bdef bfbd`, because the decoded value of `MY_FILE` is marshaled as `"��C�\u001bO���c�0��\u0026�\u0016k��[q\u0012@$éD�7��"` due to the presence of invalid UTF-8 bytes in the decoded value.

The most reasonable proposal I can think of is:

1. Stop coercing binary values to strings in the evaluator
    - This may require allowing `[]byte` values in `esc.Value`
2. Marshal all `[]byte` values as base64 (this is the standard behavior of `encoding/json`)
3. Extend the behavior of `files` so that it allows the user to supply an object with the key `data` that accepts base64-encoded data which is then decoded prior to writing the file

Another alternative for (3) is to attempt to decode all values in `files` as base64 and then use the decoded bytes if decoding succeeds. However, the likelihood of values decoding as valid base64 may be high enough that this would result in breaks in existing environments.

pgavlin:
ESC currently handles binary data by coercing binary values to a (potentially-invalid) strings. This works okay internally, since Go does not validate the wellformedness of strings when coercing from binary data. However, the standard Go JSON encoder ensures that all strings are [coerced to valid UTF-8 during marshaling](https://pkg.go.dev/encoding/json#Marshal), which replaces any invalid bytes with the replacement character. This corrupts the data.

For example, the following environment should write the bytes `dcc5 43bc 1b4f f4fb 8263 d630 a199 2688 166b c084 5b71 1240 24c3 a944 d937 e2fc` to a temporary file:

```yaml
values:
  files:
    MY_FILE:
      fn::fromBase64: 3MVDvBtP9PuCY9YwoZkmiBZrwIRbcRJAJMOpRNk34vw=
```

Instead, it writes the bytes `efbf bdef bfbd 43ef bfbd 1b4f efbf bdef bfbd efbf bd63 efbf bd30 efbf bdef bfbd 26ef bfbd 166b efbf bdef bfbd 5b71 1240 24c3 a944 efbf bd37 efbf bdef bfbd`, because the decoded value of `MY_FILE` is marshaled as `"��C�\u001bO���c�0��\u0026�\u0016k��[q\u0012@$éD�7��"` due to the presence of invalid UTF-8 bytes in the decoded value.

This commit addresses these challenges through three related changes:

1. `[]byte` values are now allowed inside of `eval.value`. This allows binary data to be faithfully represented within the evaluator.
2. `[]byte` values are now allowed inside of `esc.Value`. If an `esc.Value` contains a `[]byte`, its contents are marshaled as a base64-encoded string and a new `binary` property is set to `true`. This allows binary data to be safely roundtripped without complicated encoding schemes.
3. The `files` stanza now supports binary data. If the value of an entry in `files` is binary data, it is written to the filesystem as-is.

Strictly speaking this is a breaking change: prior to these changes, binary data--which can only be manufactured by a provider, rotator, or `fn::fromBase64`--would be marshaled as a plain old JSON string. With these changes, binary data is still marshaled as a JSON string, but its contents are now base64-encoded bytes. If users had binary data that happened to be valid UTF-8, that data will now be encoded using base64 during marshaling instead of being represented verbatim. If this is a major concern, we could change our marshaling behavior to only base64-encode if the binary data is not valid UTF-8.

Fixes #480.

### DIFF
--- a/cmd/esc/cli/cli_test.go
+++ b/cmd/esc/cli/cli_test.go
@@ -6,6 +6,7 @@ import (
 	"bytes"
 	"context"
 	"encoding/base64"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"hash/fnv"
@@ -444,7 +445,17 @@ func (c *testPulumiClient) openEnvironment(ctx context.Context, orgName, name st
 		return "", mapDiags(diags), nil
 	}
 
-	c.openEnvs[id.String()] = openEnv
+	// round-trip through JSON for better fidelity with service-backed scenarios
+	encoded, err := json.Marshal(openEnv)
+	if err != nil {
+		return "", nil, nil
+	}
+	var decoded esc.Environment
+	if err = json.Unmarshal(encoded, &decoded); err != nil {
+		return "", nil, nil
+	}
+
+	c.openEnvs[id.String()] = &decoded
 	return id.String(), mapDiags(diags), nil
 }
 
--- a/cmd/esc/cli/prepare.go
+++ b/cmd/esc/cli/prepare.go
@@ -77,15 +77,22 @@ func createTemporaryFiles(e *esc.Environment, opts PrepareOptions) (paths, envir
 
 	for _, k := range keys {
 		v := files[k]
-		s := v.Value.(string)
+
+		var contents []byte
+		switch v := v.Value.(type) {
+		case string:
+			contents = []byte(v)
+		case []byte:
+			contents = v
+		}
 
 		if v.Secret {
-			secrets = append(secrets, s)
+			secrets = append(secrets, string(contents))
 		}
 
 		path := "[unknown]"
 		if !opts.Pretend {
-			path, err = createTemporaryFile(opts.fs, []byte(s))
+			path, err = createTemporaryFile(opts.fs, contents)
 			if err != nil {
 				removeTemporaryFiles(opts.fs, paths)
 				return nil, nil, nil, err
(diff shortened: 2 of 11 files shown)


pgavlin:
https://github.com/pulumi/esc/pull/483/commits/9ab531a4e39c996880dc247967bff2a3920fbdcb is a bit questionable--might be that the better option is to change the test validation code s.t. it marshals the result of `ToJSON` before comparing.

nyobe:
I'm uneasy about having ToJSON do that base64 encoding. If it returns the underlying []byte it will end up encoding as base64 already during json.Marshal, so I think having it happen early is surprising. And I think it should be unnecessary, because the resulting Value will be able to roundtrip []byte correctly thanks to the custom MarshalJSON/UnmarshalJSON implementation.

pgavlin:
little note on the `!!binary` tag: https://yaml.org/type/binary.html

PR #488, opened 2025-03-12
Adding secret cascading to secret objects

IaroslavTitov:
### Summary
- Fixes https://github.com/pulumi/pulumi-service/issues/22804
- Adding cascading secrets to `NewSecret` method
- The root cause was that secrets do not cascade when values are retrieved from `open`s, but this is an issue for any builtin evaluators (like rotate, FromJSON)
  - I actually found a[ fix for FromJSON](https://github.com/pulumi/esc/pull/191) from awhile ago, that does the same thing, but in a less generic way. Reverting that fix now, to avoid 2 redundant places of doing the same thing.

### Testing
- After deploying pulumi-service with the fix, secret values are covered as expected and repro from the issue no longer works
- Changes to tests made in [FromJson fix](https://github.com/pulumi/esc/pull/191) still test as expected, proving that this more generic fix still works
- Added test for secret object returned from open - secretness propagates to child properties as expected


IaroslavTitov:
> Does this mean that given an object that is marked secret that contains values that are not themselves marked secret, the values can be accessed without ending up marked as secret?

Yes, exactly. See the [original issue](https://github.com/pulumi/pulumi-service/issues/22804) for a specific example using `pulumi-stacks` provider. I will also see if I can add tests specific to opens for this

eval/eval_test.go
```diff
@@ -135,7 +135,8 @@ func (testProvider) Schema() (*schema.Schema, *schema.Schema) {
 }
 
 func (testProvider) Open(ctx context.Context, inputs map[string]esc.Value, context esc.EnvExecContext) (esc.Value, error) {
-	return esc.NewValue(inputs), nil
+	_, ok := inputs["makeSecret"]
```
nyobe, line 138: I think you should either add this property to the schema, or perhaps preferably, introduce a new `testSecretProvider` that just unconditionally returns its input as a secret

eval/testdata/eval/builtin-nested-secret/env.yaml
```diff
@@ -0,0 +1,7 @@
+values:
+  open:
+    fn::open::test:
+      makeSecret: true
+      secretObject:
+        key1: secret1
+        key2: secret2
```
nyobe, line 7: I think it would be good to extend the test case to with references to the secret values to demonstrate the cascading:
```
  reference:
    # all of these should be secret
    outer: ${open}
    inner: ${open.secretObject}
    nested: ${open.secretObject.key1}
```

eval/eval_test.go
IaroslavTitov: I originally thought about just making `test-secret` provider, but figured this would be less redundant. Sounds good, I'll just make a new one

eval/testdata/eval/builtin-nested-secret/env.yaml
IaroslavTitov: Will do!

value.go
```diff
@@ -112,56 +112,41 @@ func (v *Value) UnmarshalJSON(data []byte) error {
 
 // FromJSON converts a plain-old-JSON value (i.e. a value of type nil, bool, json.Number, string, []any, or
 // map[string]any) into a Value.
-func FromJSON(v any, secret bool) (Value, error) {
-	return fromJSON("", v, secret)
+func FromJSON(v any) (Value, error) {
```
pgavlin, line 115: Why do we need this change?

IaroslavTitov: I updated the description with full explanation, but TLDR is that a year ago we fixed this exact same issue in FromJSON only, and this fix addresses it for all builtin functions, _including_ FromJSON, so there's no longer need to have it both places. The tests from that change still succeed.

pgavlin:
> > Does this mean that given an object that is marked secret that contains values that are not themselves marked secret, the values can be accessed without ending up marked as secret?
> 
> Yes, exactly. See the [original issue](https://github.com/pulumi/pulumi-service/issues/22804) for a specific example using `pulumi-stacks` provider. I will also see if I can add tests specific to opens for this

Okay, got it.

I think we may want to think through the solutions here a bit more completely:

- Should we change our APIs so that it's not possible to _create_ secret values that contain non-secret values?
- Should we change our APIs so that it's not possible to _extract_ non-secret values from secret values?

The changes in this PR are a partial implementation of the first option. These changes prevent values that are processed by `unexport` from creating secret values that contain non-secret values. I worry that this doesn't go far enough, as it's still possible to do e.g. this:

```
v := NewSecret(map[string]Value{"foo": NewValue("bar")})
```

I don't know what the right solution is there. I think it may be that the fix should be in `NewSecret`:

```
// NewSecret creates a new secret value with the given representation.
func NewSecret[T ValueType](v T) Value {
	switch v := any(v).(type) {
	case map[string]Value:
		for k, e := range v {
			if !e.Secret {
				v[k] = NewSecret(e)
			}
		}
	case []Value:
		for i, e := range v {
			if !e.Secret {
				v[i] = NewSecret(e)
			}
		}	
	}
	return Value{Value: v, Secret: true}
}
```

(though I don't know whether we should mutate the inputs or create new values)

IaroslavTitov:
I was thinking about it more in terms of your second phrasing:

> Should we change our APIs so that it's not possible to extract non-secret values from secret values?

Because it doesn't really matter what's marked secret or not inside the system, as long as all secrets are covered up on the output. So I actually originally made this same edit to `export` function, so that all values outputted would go through this cascading of secrets, but I figured that was a much higher blast radius and as far as I understand, what you're describing can only be done in a provider (but I could be completely wrong on that). I tried making a secret object just in the web console, but it enforces all secrets to be strings, not objects.

But either way, if you think adding cascading to `export` is better, I can do that.

I think editing `NewSecret()` is not ideal because:
- Secret objects can still come in from providers (so we need the cascading in unexport still)
- People can just not use the method, and manually do `Value{ secret: true }`, so we can only hope people catch that in code reviews

pgavlin:
> I was thinking about it more in terms of your second phrasing:
> 
> > Should we change our APIs so that it's not possible to extract non-secret values from secret values?
> 
> Because it doesn't really matter what's marked secret or not inside the system, as long as all secrets are covered up on the output. So I actually originally made this same edit to `export` function, so that all values outputted would go through this cascading of secrets, but I figured that was a much higher blast radius and as far as I understand, what you're describing can only be done in a provider (but I could be completely wrong on that). I tried making a secret object just in the web console, but it enforces all secrets to be strings, not objects.
> 
> But either way, if you think adding cascading to `export` is better, I can do that.
> 
> I think editing `NewSecret()` is not ideal because:
> 
> * Secret objects can still come in from providers (so we need the cascading in unexport still)
> * People can just not use the method, and manually do `Value{ secret: true }`, so we can only hope people catch that in code reviews

Changing the APIs to disallow the extraction of non-secret values from secrets would also involve changing `Value`. For example, consider a secret object that contains non-secret values:

```go
v := NewSecret(map[string]Value{"foo": NewValue("bar")})

foo := v.Value.(map[string]Value)["foo"] // foo is not secret
```

In order to avoid this, we'd need to prevent access to the underlying `map[string]Value` entirely. Which might not be bad, though it would make the API less convenient for consumers.

I'd prefer to make it more difficult to create this situation--producing values is much less common than consuming them. Hence my suggestion that we change the implementation of `NewSecret`. `Value` is already a pretty loose type; I think that it's okay for us to say "don't build values using `Value{}`, use `NewValue` and `NewSecret`, and treat `Value` values as logically immutable".

value.go
pgavlin: As a user, how do I create secret values using `FromJSON` without this parameter? Even though we added that parameter as a bug fix, I think it has independent value.

value.go
```diff
@@ -55,9 +55,44 @@ func NewValue[T ValueType](v T) Value {
...
+	case string:
+		return NewSecret(v)
+	case json.Number:
+		return NewSecret(v)
+	case bool:
+		return NewSecret(v)
+	default:
+		return Value{Value: nil, Secret: true}
+	}
+}
```
pgavlin, line 94: I would prefer that this be internal to `eval`.

For external consumers, I think it is good enough to say:
- All values should be created using `esc.NewValue`/`esc.NewSecret`, not `esc.Value{}`
- Once created, all `esc.Value`s should be treated as immutable

It's difficult to enforce those rules without significant API changes, but I think that is okay. I think that doing what you have here in the evaluator is good--it makes the evaluator safer--but I don't think that we need this for end-users (as an end-user, I don't know when or why I would use this function over `esc.NewValue`/`esc.NewSecret`).

IaroslavTitov: Updated to only have `NewSecret` as external method. I was running into generics issues (found out in go one can't cast `any` into `ValueType`, only into specific types), so I actually came up with a different way, extending `ValueType` to allow `Value` to be passed in, and just unwrapping it inside `NewSecret`, which also eliminated the need for separate `cascadeSecrets` function

value.go
IaroslavTitov: If you want the output of FromJson to be secret, can just do `esc.NewSecret(esc.FromJSON(myJson))`, which is what `evaluateBuiltinFromJSON` optionally does here:
```
		ev, err := esc.FromJSON(jv)
		if err != nil {
			e.errorf(repr.syntax(), "internal error: decoding JSON value: %v", err)
			v.unknown = true
			return v
		}

		if v.secret || ev.Secret {
			ev = esc.NewSecret(ev)
		}
		return unexport(ev, x)
```

Since secretness cannot be found in the json itself, I think it's reasonable that it comes outside like this, same as other eval methods. But up to you, I can add it back in.

value.go
```diff
@@ -28,7 +28,7 @@ import (
 
 // ValueType defines the types of concrete values stored inside a Value.
 type ValueType interface {
-	bool | json.Number | string | []Value | map[string]Value
+	bool | json.Number | string | []Value | map[string]Value | Value
```
pgavlin, line 31: I would prefer that we not add `Value` to this case, as we never store a `Value` inside of another `Value`. This type is intended as both documentation (since we can't use it as the type of `Value.Value`) and enforcement (for arguments to `NewValue` and `NewSecret`).

Instead, how about a method on `Value`--say `MakeSecret`--that copies a value into a new, secret value?

```go
func (v Value) MakeSecret() Value {
	switch v := any(v.Value).(type) {
	case bool:
		return NewSecret(v)
	case json.Number:
		return NewSecret(v)
	case string:
		return NewSecret(v)
	case []Value:
		return NewSecret(v)
	case map[string]Value:
		return NewSecret(v)
	default:
		panic("invalid value")
	}
}
```

IaroslavTitov: This sounds good, reverted back to 2 method solution, this is simpler/more elegant than my original `CascadeSecrets`, and achieves the same thing

pgavlin (approved):
LGTM! Thanks for seeing this one through.

PR #491, opened 2025-03-13
Don't decrypt secrets from invalid top-level keys

seanyeh:
We should only decrypt secrets under `values`. This fixes decryption errors when we have invalid ciphertexts under invalid top-level keys (which we don't validate for)

### DIFF
--- a/eval/crypt.go
+++ b/eval/crypt.go
@@ -23,6 +23,7 @@ import (
 	"fmt"
 	"hash/crc32"
 	"io"
+	"strings"
 
 	"github.com/pulumi/esc/syntax"
 	"github.com/pulumi/esc/syntax/encoding"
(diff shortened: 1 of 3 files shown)


eval/crypt.go
```diff
@@ -150,6 +151,12 @@ func DecryptSecrets(ctx context.Context, filename string, source []byte, decrypt
 			return n, nil, nil
 		}
 
+		// Don't attempt to decrypt secrets outside of valid top-level keys
+		path := n.Syntax().Path()
+		if !(strings.HasPrefix(path, "values.") || strings.HasPrefix(path, "imports.")) {
```
seanyeh, line 156: @pgavlin thoughts on this approach? another approach is to change where we start walking through the nodes (only walking through `values/imports`) but that's more involved since that logic is shared between encrypt/decrypt

pgavlin: I think that this is okay. I don't love that we're using a string for the path here--ideally the path would be stored in a parsed form--but this is essentially what I'd suggest.

The other similar option is to change `rewriteYAML` so that it passes the current path to the visitor. Something like

```
func rewriteYAML(
	ctx context.Context,
	filename string,
	source []byte,
	visitor func(path []any, n syntax.Node) (syntax.Node, syntax.Diagnostics, error),
) ([]byte, error) {
...
}
```

But that would also require changing `syntax.Walk`. So I think what you have is fine.

I'd also go a little further and only rewrite secrets under `values`.

pgavlin (approved):

PR #504, opened 2025-03-28
fix rotate paths roundtripping

nyobe:
As part of adding the [retry-rotation button](https://github.com/pulumi/pulumi-service/pull/27132) to the console, I realized that the rotation paths that we output as part of rotation results cannot be directly used as rotation path inputs.  This is because rotation path inputs get implicitly prefixed by `values.`, while the result paths are physical docpaths that already have the `values` prefix.

Right now the frontend works around this by strip the `values.?` prefix from the result paths, but this just exposes the fact that the prefixing doesn't actually work correctly for quoted keys like `["rotated-creds"] -> values.["rotated-creds"] -> invalid path`

I think that doing this prefixing in eval.RotateEnvironment is probably a mistake. It was done to try to match the behavior of the paths used by the CLI. So, while this is technically a breaking change, I propose moving the prefixing behavior to the CLI instead, and have the eval apis just use physical docpaths consistently.

### DIFF
--- a/cmd/esc/cli/env_rotate.go
+++ b/cmd/esc/cli/env_rotate.go
@@ -44,6 +44,7 @@ func newEnvRotateCmd(envcmd *envCommand) *cobra.Command {
 				if err != nil {
 					return fmt.Errorf("'%s' is an invalid property path: %w", arg, err)
 				}
+				// implicitly prefix the given path with the values top level key
 				path = slices.Insert(path, 0, "values")
 				rotationPaths = append(rotationPaths, path.String())
 			}
--- a/cmd/esc/cli/env_rotate.go
+++ b/cmd/esc/cli/env_rotate.go
@@ -5,6 +5,7 @@ package cli
 import (
 	"context"
 	"fmt"
+	"slices"
 	"strings"
 
 	"github.com/pulumi/esc/cmd/esc/cli/client"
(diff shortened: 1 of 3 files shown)


seanyeh (approved):

IaroslavTitov (commented):
Would this break anything in pulumi-service? Probably have to manually update to the new version and fix up where you add `values.` in the console?

IaroslavTitov (approved):

nyobe:
> Would this break anything in pulumi-service? Probably have to manually update to the new version and fix up where you add `values.` in the console?

yeah, we'll need to update the frontend at the same time as bumping the ESC version to back out [this workaround](https://github.com/pulumi/pulumi-service/pull/27132/files/d5b7ca39ebbe4f858332f2a75ef34dd418e03e19#r2015454293)

I don't believe we're inputting rotation paths anywhere else.  The biggest concern I have is that the `rotate` command in the  current version of the CLI won't send the paths with the values prefix to the API, so it won't work correctly.  But I suppose we could add a check in the rotate api handler to see if paths are prefixed or not, and try to fix them up if not.

IaroslavTitov:
> I suppose we could add a check in the rotate api handler to see if paths are prefixed or not, and try to fix them up if not.

Yeah, def need to make sure this doesn't break any of the callers, like frontend, CLI, ESC SDK

pgavlin (approved):

PR #510, opened 2025-04-21
Enable client lookup of default orgs if not locally set

caseyyh:
## Context

Fix https://github.com/pulumi/pulumi-service/issues/27575

A goal of Project Hockeystick is to have new Pulumi users to be able to trial premium Pulumi features immediately; as a part of this, we would want new Pulumi users to have their `esc` usage directly hook into to the trial org with Premium features enabled, rather than their individual org.

For details on the project, see the epic: https://github.com/pulumi/home/issues/3910

## Change

1. Adds integration against the new `GetDefaultOrg` API call: https://github.com/pulumi/pulumi-service/pull/27461
2. Adds fallback logic to make the API call if `~/.pulumi/config` does not have default org set for the given backend;
3. Move fallback logic of falling back to the individual org to after the API call is made.

### DIFF
--- a/cmd/esc/cli/esc.go
+++ b/cmd/esc/cli/esc.go
@@ -179,25 +179,27 @@ func (esc *escCommand) confirmPrompt(prompt, name string) bool {
 }
 
 // Looks up the default org.
-// Prefers default org that the user has configured locally in their ~/.pulumi/config.
+// Prefers default org that the user has configured locally in their ~/.pulumi/config.json
 // If unset, then it will attempt to make an API call to the backend to determine the service's opinion
 // of which user organization should be the default; defaults to individual org otherwise if unset.
 func (esc *escCommand) lookupDefaultOrg(ctx context.Context, backendURL, username string) (string, error) {
 	userConfiguredDefaultOrg, err := esc.workspace.GetBackendConfigDefaultOrg(backendURL, username)
-	if err != nil || userConfiguredDefaultOrg != "" {
-		return userConfiguredDefaultOrg, err
-	}
-
-	// If client is unset, return the individual org.
-	if esc.client == nil {
-		return username, nil
-	}
-
-	backendDefaultOrg, err := esc.client.GetDefaultOrg(ctx)
 	if err != nil {
-		return backendDefaultOrg, err
-	} else if backendDefaultOrg == "" {
-		return username, nil
+		return "", err
 	}
-	return backendDefaultOrg, err
+	if userConfiguredDefaultOrg != "" {
+		return userConfiguredDefaultOrg, nil
+	}
+
+	if esc.client != nil {
+		backendDefaultOrg, err := esc.client.GetDefaultOrg(ctx)
+		if err != nil {
+			return backendDefaultOrg, err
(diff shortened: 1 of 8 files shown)


cmd/esc/cli/login.go
```diff
@@ -172,14 +172,8 @@ func (esc *escCommand) getCachedClient(ctx context.Context) error {
...
 		}
 
-		defaultOrg, err := esc.workspace.GetBackendConfigDefaultOrg(backendURL, nAccount.Username)
-		if err != nil {
-			return fmt.Errorf("could not determine default org: %w", err)
-		}
-
 		account = &workspace.Account{
-			Account:    *nAccount,
-			DefaultOrg: defaultOrg,
```
caseyyh, line 182: From reading the code, it seems like this (and L212-216) are both safe to remove.

`defaultOrg` configuration appears to get overridden later by `getCachedCredentials`, anyway. I've moved the default org setting to be after the client is initialized.

cmd/esc/cli/esc.go
```diff
@@ -176,3 +177,27 @@ func (esc *escCommand) confirmPrompt(prompt, name string) bool {
 	line, _ := reader.ReadString('\n')
 	return strings.TrimSpace(line) == name
 }
+
+// Looks up the default org.
+// Prefers default org that the user has configured locally in their ~/.pulumi/config.
```
IaroslavTitov, line 182:
```suggestion
// Prefers default org that the user has configured locally in their ~/.pulumi/config.json
```

cmd/esc/cli/esc.go
```diff
@@ -176,3 +177,27 @@ func (esc *escCommand) confirmPrompt(prompt, name string) bool {
...
+// If unset, then it will attempt to make an API call to the backend to determine the service's opinion
+// of which user organization should be the default; defaults to individual org otherwise if unset.
+func (esc *escCommand) lookupDefaultOrg(ctx context.Context, backendURL, username string) (string, error) {
+	userConfiguredDefaultOrg, err := esc.workspace.GetBackendConfigDefaultOrg(backendURL, username)
+	if err != nil || userConfiguredDefaultOrg != "" {
```
IaroslavTitov, line 187: nit: This line is a bit confusing, took me a second to understand. I think it would be better to split into 2 if's for readability, but up to you

IaroslavTitov: Actually, looking more at this function, I'd prefer something like this to make the function easier to read/understand
```
	userConfiguredDefaultOrg, err := esc.workspace.GetBackendConfigDefaultOrg(backendURL, username)
	if err != nil {
		return "", err
	}
	if userConfiguredDefaultOrg != "" {
		return userConfiguredDefaultOrg, nil
	}

	if esc.client != nil {
		backendDefaultOrg, err := esc.client.GetDefaultOrg(ctx)
		if err != nil {
			return "", err
		}
		if backendDefaultOrg != "" {
			return backendDefaultOrg, nil
		}
	}

	return username, nil
```

But totally up to you, this is just a codestyle nit

IaroslavTitov (approved):
LGTM, a few nit comments, no blockers

PR #512, opened 2025-04-24
syntax: add a node for document trivia

### LINKED ISSUE #260: Top level comments are dropped during auto-formatting
### What happened?

I was trying to comment my a env I was creating, and my comment got eaten. Experimentation shows that the comment only persists if it is attached to an item.

### Example

https://github.com/pulumi/esc/assets/22222529/58d9c34f-f487-4528-80c5-0056ab863e42

### Output of `pulumi about`

NA

### Additional context

_No response_

### Contributing

pgavlin:
These changes add a new node type, DocumentNode, to the syntax package. This node allows concrete syntaxes to attach document-level syntactical trivia (e.g. comments, metadata, etc.) that is not otherwise associated with a node.

These changes also replace `*RotationResult` with `RotationResult`, as the result is already nilable. This fixes an issue when running the eval tests with `PULUMI_ACCEPT` set.

Fixes #260.

### DIFF
--- a/syntax/nodes.go
+++ b/syntax/nodes.go
@@ -342,3 +342,48 @@ func (n *ObjectNode) String() string {
 	}
 	return fmt.Sprintf("{ %s }", strings.Join(s, ", "))
 }
+
+// A DocumentNode represents a complete document. A DocumentNode wraps a single nodes and mostly exists as a holder for
+// additional syntactical trivia or metadata (e.g. comments). A DocumentNode should only appear at the root of a syntax tree.
+type DocumentNode struct {
+	node
+
+	content Node
+}
+
+// DocumentSyntax creates a new document with the given contents and associated syntax.
+func DocumentSyntax(syntax Syntax, content Node) *DocumentNode {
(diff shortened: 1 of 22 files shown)


syntax/nodes.go
```diff
@@ -342,3 +342,48 @@ func (n *ObjectNode) String() string {
 	}
 	return fmt.Sprintf("{ %s }", strings.Join(s, ", "))
 }
+
+// A DocumentNode represents a complete document. A DocumentNode wraps a single nodes and mostly exists as a holder for
```
borisschlosser, line 346:
```suggestion
// A DocumentNode represents a complete document. A DocumentNode wraps a single node and mostly exists as a holder for
```

borisschlosser (commented):
lgtm! Still asking others who are more familiar with the code to approve.

nyobe (approved):

pgavlin:
So I haven't merged this yet because I'm not totally sure that it's the right approach.

Part of me feels like this is shoehorning a YAML concept into the generic syntax layer, and that a better approach would be to somehow capture this in `YAMLSyntax`. Any thoughts here?

PR #513, opened 2025-04-24
esc: make traces optional

pgavlin:
And add a couple of helper methods: one to remove trace metadata from values and one to remove all metadata from an Environment. This allows embedders to more easily provide less heavyweight representations of evaluated values.

### DIFF
--- a/cmd/esc/cli/testdata/open-detailed.yaml
+++ b/cmd/esc/cli/testdata/open-detailed.yaml
@@ -68,20 +68,6 @@ environments:
         }
       }
     }
-  },
-  "trace": {
-    "def": {
-      "begin": {
-        "line": 0,
-        "column": 0,
-        "byte": 0
-      },
-      "end": {
-        "line": 0,
-        "column": 0,
-        "byte": 0
-      }
-    }
   }
 }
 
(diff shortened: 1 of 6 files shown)


pgavlin:
Part of the motivation here is the potential to add support in the ESC API for returning an opened environment in a form that does not include this metadata. The metadata-free form is likely to be easier to consume for most API-level users.

borisschlosser (approved):
lgtm! Let me know your thoughts where we can leverage these changes in ESC.

nyobe:
I think rather than stripping metadata, it might be nice to have a helper that converts an esc.Value tree into a go value. I think users asking for a simpler representation probably just want a direct mapping to the shape the environment defines. 

Then we don't lossily drop metadata, we just provide a convenience mapping for direct lookups when desired.

pgavlin:
> Edit: actually we already have this with the ToJSON() helper!

Yes--we already have `ToJSON`. This is intended to allow us to do something similar to `ToJSON` at the REST API level but maintain secret-ness metadata.

It could be that we should just add a parameter to the REST API that allows _all_ metadata to be discarded (i.e. call `ToJSON` and return the result). I just don't love dropping the secret-ness.

PR #487, opened 2025-03-12
chore: freeze 0.13.0

nyobe:

PR #514, opened 2025-04-24
Codefreeze for 0.13.1

IaroslavTitov:
### Summary
- Codefreeze for 0.13.1


CHANGELOG_PENDING.md
```diff
@@ -1,15 +1,5 @@
 ### Improvements
 
-- Updated to go 1.23
```
nyobe, line 3: hold off on removing the changelog pending entries until the following commit, because its what is used for the release notes: https://github.com/pulumi/esc/pull/487#discussion_r1992040873

IaroslavTitov: Ah, good catch, removed changes to pending changelog

nyobe (approved):

PR #546, opened 2025-06-06
Always expect environment name as first argument for run 

### LINKED ISSUE #478: env run not disambiguating args correctly
### What happened?

I was trying to use `pulumi env run` (which just uses esc under the hood), but forgot to set the environment name in the command line args.  This resulted in a rather confusing error.

### Example

```
$ pulumi env run -- pulumi preview                                                                        
Warning: Referring to an environment name ('pulumi') without a project is deprecated.
Please use 'v-thomas-pulumi-corp/default/pulumi' or 'default/pulumi' instead.
error: resolving command: exec: "preview": executable file not found in $PATH
```

With the `--` disambiguation I was expecting it to warn me that there was no environment name specified on the command line.

### Output of `pulumi about`

```
$ pulumi about                    
CLI          
Version      3.153.2-dev.0
Go Version   go1.22.1
Go Compiler  gc

Plugins
KIND      NAME           VERSION
resource  lambda-plugin  0.0.1
language  python         3.153.2-dev.0

Host     
OS       debian
Version  trixie/sid
Arch     x86_64
```

### Additional context

_No response_

### Contributing

tehsis:
Fixes #478 

This PR checks ensures that the first argument for _run_ is always an environment. This is required for when `--` is used since in the current behavior will threat the first argument after `--` as the environment.

**Before:**
image

**After:**
image

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -1,5 +1,8 @@
 ### Improvements
 
+- `esc run` will expect environment to be pased before `--`
+  [#545](https://github.com/pulumi/esc/pull/545)
+
 ### Bug Fixes
 
 - Fix `esc version`
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -1,7 +1,7 @@
 ### Improvements
 
 - `esc run` expects environment to be passed before `--`
-  [#545](https://github.com/pulumi/esc/pull/545)
+  [#545](https://github.com/pulumi/esc/pull/546)
 
 ### Bug Fixes
 
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -1,6 +1,6 @@
 ### Improvements
 
-- `esc run` will expect environment to be pased before `--`
+- `esc run` expects environment to be passed before `--`
   [#545](https://github.com/pulumi/esc/pull/545)
 
 ### Bug Fixes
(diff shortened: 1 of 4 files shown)


cmd/esc/cli/env_run.go
```diff
@@ -127,19 +127,29 @@ func newEnvRunCmd(envcmd *envCommand) *cobra.Command {
 				return err
 			}
 
-			ref, args, err := envcmd.getExistingEnvRef(ctx, args)
-			if err != nil {
-				return err
+			// If -- was used but there's no arguments, it means no environment was specified before the command
+			if len(args) == 0 || cmd.ArgsLenAtDash() == 0 {
```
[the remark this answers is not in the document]
tehsis: I think a better solution would be to print the help on errors, but we should cover that on a different issue.

cmd/esc/cli/env_run.go
```diff
@@ -127,19 +127,29 @@ func newEnvRunCmd(envcmd *envCommand) *cobra.Command {
...
 			}
 
-			if len(args) == 0 {
+			if len(args) == 1 {
 				return fmt.Errorf("no command specified")
 			}
-			command, err := envcmd.esc.exec.LookPath(args[0])
+
+			// First argument is always the environment name
+			envRefArgs := args[:1]
```
seanyeh, line 140: Is there a reason to split the args like this, or is it just for clarity?? I expect the existing code here still works and doesn't need any changes

CHANGELOG_PENDING.md
```diff
@@ -1,5 +1,8 @@
 ### Improvements
 
+- `esc run` expects environment to be passed before `--`
+  [#545](https://github.com/pulumi/esc/pull/545)
```
seanyeh, line 4: 546

cmd/esc/cli/env_run.go
tehsis: Reverted back, Originally I was doing more parsing on the env name for improved error reporting but I think that's better to be handled on a different pr.

PR #552, opened 2025-06-26
[approvals] Add draft flag to env set, edit, version rollback

rgharris:
Automatically submit change request for simplicity. The console does the same currently.

The flag is hidden for now while the feature and APIs are in preview.

Closes https://github.com/pulumi/pulumi-service/issues/28046

Examples:
```
> esc env set example/g values.example abc2 --draft
Change request created: a9b6ab9b-54b6-4ad6-8236-0d0d1e4eba48
Change request URL: https://app.pulumi.com/pulumipus-org/esc/example/g/change-requests/a9b6ab9b-54b6-4ad6-8236-0d0d1e4eba48
Change request submitted

> esc env edit example/g --draft
Change request created: bc15bae8-2afb-4a63-8d50-ffefef7fbe77
Change request URL: https://app.pulumi.com/pulumipus-org/esc/example/g/change-requests/bc15bae8-2afb-4a63-8d50-ffefef7fbe77
Change request submitted

> esc env version rollback example/g@1 --draft
Change request created: a0f6c39d-c249-44c4-b9e7-2cdf26397705
Change request URL: https://app.pulumi.com/pulumipus-org/esc/example/g/change-requests/a0f6c39d-c249-44c4-b9e7-2cdf26397705
Change request submitted
```

### DIFF
--- a/cmd/esc/cli/client/client.go
+++ b/cmd/esc/cli/client/client.go
@@ -170,10 +170,12 @@ type Client interface {
 	) (string, []EnvironmentDiagnostic, error)
 
 	// SubmitChangeRequest submits the change request with the specified ID in org orgName.
+	// Optionally provide a description for the change request.
 	SubmitChangeRequest(
 		ctx context.Context,
 		orgName string,
 		changeRequestID string,
+		description *string,
 	) error
 
 	// DeleteEnvironment deletes the environment envName in org orgName.
@@ -686,8 +688,9 @@ func (pc *client) SubmitChangeRequest(
 	ctx context.Context,
 	orgName string,
 	changeRequestID string,
+	description *string,
 ) error {
-	req := SubmitChangeRequestRequest{}
+	req := SubmitChangeRequestRequest{Description: description}
 	path := fmt.Sprintf("/api/preview/change-requests/%v/%v/submit", orgName, changeRequestID)
 	err := pc.restCall(ctx, http.MethodPost, path, nil, &req, nil)
 	return err
--- a/cmd/esc/cli/client/client.go
+++ b/cmd/esc/cli/client/client.go
@@ -152,6 +152,30 @@ type Client interface {
 		etag string,
 	) ([]EnvironmentDiagnostic, error)
 
+	// CreateEnvironmentDraft creates a draft update of the YAML for the environment envName in org orgName.
+	// The resulting ChangeRequestID is returned.
+	//
+	// If the new environment definition contains errors, the creation will fail with diagnostics.
+	//
+	// If etag is not the empty string and the environment's current etag does not match the provided etag
+	// (i.e. because a different entity has called UpdateEnvironment), the creation will fail with a 409
+	// error.
+	CreateEnvironmentDraft(
+		ctx context.Context,
+		orgName string,
+		projectName string,
+		envName string,
+		yaml []byte,
+		etag string,
+	) (string, []EnvironmentDiagnostic, error)
+
+	// SubmitChangeRequest submits the change request with the specified ID in org orgName.
+	SubmitChangeRequest(
+		ctx context.Context,
+		orgName string,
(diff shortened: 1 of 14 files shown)


cmd/esc/cli/client/client.go
```diff
@@ -152,6 +152,30 @@ type Client interface {
...
+		envName string,
+		yaml []byte,
+		etag string,
+	) (string, []EnvironmentDiagnostic, error)
+
+	// SubmitChangeRequest submits the change request with the specified ID in org orgName.
+	SubmitChangeRequest(
+		ctx context.Context,
+		orgName string,
+		changeRequestID string,
```
nyobe, line 176: should description be an optional argument for the client?

cmd/esc/cli/client/client_test.go
```diff
@@ -387,6 +387,151 @@ func TestUpdateEnvironment(t *testing.T) {
 	})
 }
 
+func TestCreateEnvironmentDraft(t *testing.T) {
```
nyobe, line 390: tbh I think you can drop most of these tests beyond the first one, because they are mostly just exercising the mock.  the main interesting thing seems to be that the tag argument is propagated as a header?

cmd/esc/cli/env_edit.go
```diff
@@ -74,13 +75,32 @@ func newEnvEditCmd(env *envCommand) *cobra.Command {
...
+					}
+				} else {
+					diags, err = edit.env.esc.client.UpdateEnvironmentWithProject(ctx, ref.orgName, ref.projectName, ref.envName, yaml, "")
+					if err != nil {
+						return fmt.Errorf("updating environment definition: %w", err)
+					}
+					if len(diags) == 0 {
+						fmt.Fprintln(edit.env.esc.stdout, "Environment updated.")
+						return nil
+					}
```
nyobe, line 103: this is repeated three times, I wonder if it makes sense to pull it into a function ?

nyobe (approved):
lgtm!  I think it'd be really neat to also support a `--draft=<change request id>` flag for updating or reading a draft version

cmd/esc/cli/client/client.go
rgharris: Added in edcbe7d, let me know if you like the approach

cmd/esc/cli/env_edit.go
rgharris: Good call, cleaned up in 3e0c26c. I previously assumed it would be clunky to extract but worked out just fine. I also added some minimal CLI tests in 0b8ac32 and 25270a2.

nyobe (approved):

cmd/esc/cli/env.go
```diff
@@ -389,3 +389,55 @@ func (cmd *envCommand) writePropertyEnvironmentDiagnostics(out io.Writer, diags
...
+		}
+		return diags, nil
+	} else {
+		diags, err := esc.client.UpdateEnvironmentWithProject(ctx, ref.orgName, ref.projectName, ref.envName, yaml, tag)
+		if err != nil {
+			return nil, fmt.Errorf("updating environment definition: %w", err)
+		}
+		if len(diags) == 0 && envUpdateSuccessMessage != "" {
+			fmt.Fprintln(esc.stdout, envUpdateSuccessMessage)
+		}
```
nyobe, line 440: I think you could maybe make this output the responsibility of the caller? 🤔

nyobe (approved):

cmd/esc/cli/env.go
rgharris: The reason I didn't is because all of them want to print the draft/CR output. So we'd need to specifically check if `!draft` and then log the message in each caller if no `diags`. Either way it's a bit clunky, can refactor to remove from here if you think it's best?

PR #556, opened 2025-07-16
Added support for passing values read from a file using --file

tehsis:
# Description

This PR adds supports for `--file <path>` modifier in `esc env set`. 
This modifier will read the file in <path> and, if it's a valid string content, will use that to set the value of the specified key.

## Usage

```sh
$ esc env set --file /path/to/file.txt myorg/myproj/myenv "mykey"
```

the `--file` flag also supports passing "-" to read from `stdin` so its consistent with the behavior of `--file` flag in `esc env edit`

```sh
$ cat /path/to/file.txt | esc env set --file - myorg/myproj/myenv
```

The flag can be combined with other flags. For example to store a file as a rawstring content and secret:

```sh
$ esc env set --secret --string --file /path/to/file.pem myorg/myproj/myenv "mykey"
```

To simplify reading file content, we fixed a bug where a new line is always being added at the end of the value when using `--value string` in `esc env open` and `esc env get`. 

Users will be able to get the file content from esc.

eg.

```sh
$ esc env set --secret --string --file /path/to/file.pem myorg/myproj/myenv "mypemfile"
$ esc env get --show-secrets --value string myorg/myproj/myenv "mypemfile" >> /path/to/file_1.pem
$ diff /path/to/file_1.pem /path/to/file.pem
$
```

## Known Issues

`--file` does not currently supports binary files (or any other file that does not contain valid utf-8 characters). These files require special treatment and might be included in another change (see #483)

### DIFF
--- a/cmd/esc/cli/env_set.go
+++ b/cmd/esc/cli/env_set.go
@@ -72,7 +72,9 @@ func newEnvSetCmd(env *envCommand) *cobra.Command {
 				switch file {
 				case "-":
 					content, err = io.ReadAll(env.esc.stdin)
-					break
+					if err != nil {
+						return fmt.Errorf("could not read from stdin: %w", err)
+					}
 				default:
 					content, err = os.ReadFile(file)
 					if err != nil {
--- a/cmd/esc/cli/env_set.go
+++ b/cmd/esc/cli/env_set.go
@@ -72,6 +72,7 @@ func newEnvSetCmd(env *envCommand) *cobra.Command {
 				switch file {
 				case "-":
 					content, err = io.ReadAll(env.esc.stdin)
+					break
 				default:
 					content, err = os.ReadFile(file)
 					if err != nil {
--- a/cmd/esc/cli/env_set.go
+++ b/cmd/esc/cli/env_set.go
@@ -5,9 +5,11 @@ package cli
 import (
 	"context"
 	"fmt"
+	"os"
 	"regexp"
 	"strconv"
 	"strings"
+	"unicode/utf8"
 
 	"github.com/ccojocar/zxcvbn-go"
 	"github.com/spf13/cobra"
@@ -23,6 +25,7 @@ func newEnvSetCmd(env *envCommand) *cobra.Command {
 	var plaintext bool
 	var rawString bool
 	var draft bool
+	var fromFile bool
 
 	cmd := &cobra.Command{
 		Use:   "set [<org-name>/][<project-name>/]<environment-name> <path> <value>",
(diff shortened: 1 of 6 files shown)


cmd/esc/cli/env_set.go
```diff
@@ -47,19 +51,45 @@ func newEnvSetCmd(env *envCommand) *cobra.Command {
 			if ref.version != "" {
 				return fmt.Errorf("the set command does not accept versions")
 			}
-			if len(args) < 2 {
+
+			if len(args) < 2 && file == "" {
 				return fmt.Errorf("expected a path and a value")
 			}
```
pgavlin, line 57: probably also want to validate that `len(args) == 1` if `file` is specified

cmd/esc/cli/env_set.go
```diff
@@ -47,19 +51,45 @@ func newEnvSetCmd(env *envCommand) *cobra.Command {
...
+			if file != "" {
+				var content []byte
+				switch file {
+				case "-":
+					content, err = io.ReadAll(env.esc.stdin)
+					if err != nil {
+						return fmt.Errorf("could not read from stdin: %w", err)
+					}
+				default:
+					content, err = os.ReadFile(file)
```
pgavlin, line 78: this needs to use `env.esc.fs` in order to maintain testability

pgavlin (changes requested):
These changes need tests. We have a pretty robust test harness for the CLI's code--you can use [the tests for `env set`](https://github.com/pulumi/esc/blob/main/cmd/esc/cli/testdata/env-set.yaml) and [the tests for `env edit -f`](https://github.com/pulumi/esc/blob/main/cmd/esc/cli/testdata/env-edit-file.yaml) for reference. I'd recommend adding tests in a new testdata file (e.g. `env-set-file.yaml`).

cmd/esc/cli/testdata/env-set-file.yaml
```diff
@@ -0,0 +1,80 @@
...
+  esc env get default/test
+  esc env get default/test --value string
+  esc env set default/test foo -f=test.file --string
+  esc env get default/test
+  esc env set default/test -f=test.file || echo -n ""
+  esc env set -f=test.file || echo -n ""
+environments:
+  test-user/default/test:
+    values:
+      foo: bar
```
pgavlin, line 15: we could use a test for invalid UTF-8 in both stdin and a file on disk. you _should_ be able to use the `process` property to add files that are accessible to the commands in `run`:

```suggestion
  echo "hello" | esc env set default/test foo -f=-
  esc env get default/test
  echo "hello world" >test.file
  esc env set default/test foo -f=test.file
  esc env get default/test
  esc env get default/test --value string
  esc env set default/test foo -f=test.file --string
  esc env get default/test
  esc env set default/test -f=test.file || echo -n ""
  esc env set -f=test.file || echo -n ""
  esc env set default/test -f=binary foo || echo -n ""
  esc env set -f=- default/test foo <binary || echo -n ""
environments:
  test-user/default/test:
    values:
      foo: bar
process:
  fs:
    binary: !!binary sFOZJGpvmbYHwHRLnCYSnQ==
```

cmd/esc/cli/testdata/env-set-file.yaml
```diff
@@ -0,0 +1,80 @@
+run: |
+  echo "hello" | esc env set default/test foo -f=-
+  esc env get default/test
+  echo "hello world" >test.file
+  esc env set default/test foo -f=test.file
+  esc env get default/test
+  esc env get default/test --value string
+  esc env set default/test foo -f=test.file --string
```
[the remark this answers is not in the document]
tehsis: I've added tests for this

pgavlin (approved):

Issue #557, opened 2025-07-18
Decryption error with keys with dashes


seanyeh:
1. Create a secret for a key containing dashes (e.g. `x-`).
2. `esc env get <env> --show-secrets` fails with the following:
```
Error: getting environment metadata: [0]
Diags: decrypting: static secrets are not supported in anonymous environments
```

PR #559, opened 2025-07-18
Fix invalid toplevel key check

seanyeh:
Related to https://github.com/pulumi/esc/issues/557

We previously added a check to avoid decrypting from invalid top-level keys (on check env), but this also affected valid keys with symbols (due to the use of bracket syntax)

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -1,13 +1,8 @@
 ### Improvements
 
-- `esc run` expects environment to be passed before `--`
-  [#545](https://github.com/pulumi/esc/pull/546)
-- `esc env set` uses a more readable YAML format when setting a key in an empty map  
-  [#548](https://github.com/pulumi/esc/pull/548)
-
 ### Bug Fixes
 
-- Fix `esc version`
- [#541](https://github.com/pulumi/esc/pull/541)
+- Fix decryption error with keys with dashes
+ [#559](https://github.com/pulumi/esc/pull/559)
 
 ### Breaking changes


eval/crypt.go
```diff
@@ -153,7 +153,7 @@ func DecryptSecrets(ctx context.Context, filename string, source []byte, decrypt
 
 		// Don't attempt to decrypt secrets outside of "values"
 		path := n.Syntax().Path()
-		if !strings.HasPrefix(path, "values.") {
+		if !strings.HasPrefix(path, "values.") && !strings.HasPrefix(path, "values[") {
```
seanyeh, line 156: @pgavlin this works, but do we have a nicer way of checking if something is within the toplevel `values`?

pgavlin: We could inspect the structure of the document in `rewriteYAML` and only rewrite the value of the `values` key, if any exists. Something like

```go
if obj, ok := syn.(*syntax.ObjectNode); ok {
	for i := 0; i < obj.Len(); i++ {
		kvp := obj.Index(i)
		if kvp.Key.Value() == "values" {
			v, diags, err := syntax.Walk(kvp.Value)
			if err != nil {
				return nil, err
			}
			if len(diags) != 0 {
				return nil, diags
			}
			kvp.Value = v
			obj.SetIndex(i, kvp)
			break
		}
	}
}
```

seanyeh: the issue with that is we also use `rewriteYAML` on encrypt and i want that to rewrite the whole doc (in case someone accidentally puts a secret in an invalid top-level) - which maintains current behavior

I've updated this PR to use a more elegant check using property paths

IaroslavTitov (approved):
The change looks fine, but I'm curious how example works - why does having a dash in a key break check of root `values` key?

seanyeh:
@IaroslavTitov paths are usually represented in dot notation but those that include certain symbols are represented in bracket syntax (think `values["key-"]` vs. `values.key`). Using property paths rather than string matching is more correct here

nyobe (approved):

PR #563, opened 2025-07-28
goreleaser: enable pre-release releases

nyobe:
Enable goreleaser to publish pre-release artifacts

### DIFF
--- a/.goreleaser.prerelease.yml
+++ b/.goreleaser.prerelease.yml
@@ -3,7 +3,7 @@ project_name: esc
 changelog:
   skip: true
 release:
-  disable: true
+  disable: false
 snapshot:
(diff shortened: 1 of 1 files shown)


pgavlin:
ah, different browser worked. so does this mean that goreleaser will only create prereleases? or will it always mark releases as prerelease?

nyobe:
it's in the `.goreleaser.prerelease.yml` so it should only be used for the [pre-release pipeline](https://github.com/pulumi/esc/blob/main/.github/workflows/publish-prerelease.yaml).

that said I think the `prerelease` flag is wrong, it [looks like it's supposed to determine this from the tag](https://github.com/goreleaser/goreleaser/pull/871) w/o additional configuration

pgavlin (approved):

PR #569, opened 2025-07-31
[Approvals] Use non preview routes

rgharris:
- [x] Merge this only after testing with the non preview routes in prod.

Closes pulumi/pulumi-service#30757

We don't want these to be used by the CLI when the feature moves to GA in a PR that will be merged later.


rgharris:
I've tested this with prod after the needed API change was shipped this morning.

PR #572, opened 2025-08-01
chore: freeze v0.16.1

rgharris:


rgharris:
> do you want to unhide the flag as part of this?

@nyobe I figured I could ship this quickly before we move to GA then do a minor version for GA? Happy to hold it though if preferred.

rgharris:
Closing this as we're going to ship a version with the following right away:
- https://github.com/pulumi/esc/pull/566

PR #577, opened 2025-08-07
eval: prevent infinite recursion in export cycle

pgavlin:
These changes guard against infinte recursion when exporting values. The
export process should never encounter cycles, but if we do, we would
rather produce partial results than overflow the stack. These changes
detect cycles during export and replace the in-progress value with an
unknown value. Future changes should issue an internal error in this
case.


nyobe (approved):
lets add some telemetry in a followup, since this shouldn't be possible

PR #581, opened 2025-09-01
Enforce reserved properties to be non-null objects

borisschlosser:
This PR enforces [reserved properties](https://www.pulumi.com/docs/esc/reference/reserved-properties/) to be non-null objects.

Previously only `imports` and `values` have been enforced to be non-null objects. Now also reserved properties are enforced to be non-null objects.

This example environment now errors:
``` 
values:
  foo: bar
  pulumiConfig: null
```

### DIFF
--- a/ast/environment.go
+++ b/ast/environment.go
@@ -27,6 +27,9 @@ import (
 	"github.com/pulumi/pulumi/sdk/v3/go/common/util/contract"
 )
 
+// See https://www.pulumi.com/docs/esc/reference/reserved-properties/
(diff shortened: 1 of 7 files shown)


pgavlin (changes requested):
I don't think we should take this approach.

These reserved properties are not characteristics of the ESC language/evaluator itself. They are defined by the layers that interpret evaluated ESC environments. IMO we should handle this at that layer.

One approach that seems promising would be to allow the evaluator to accept an output schema that can be used to validate the result of evaluation. We could use that schema to perform richer validation as well as this sort of check.

cc @nyobe, @seanyeh

Issue #287, opened 2024-04-16
Add fn::concat support


kpitzen:
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details

I would like to have the ability to combine arrays in a given environment.  

One could imagine one environment, `test-0` with the following config:
```yaml
  pulumiConfig:
    examples-0:
      - foo
      - bar
```

and a second environment, `test-1` which imports `test-0` and would like to do the following:
```yaml
  pulumiConfig:
    examples-1:
      -baz
    examples:
      fn::concat:
        - ${pulumiConfig.examples-0}
        - ${pulumiConfig.examples-1}
```
Where the resulting config would be rendered as:
```yaml
  pulumiConfig:
    examples-1:
      -baz
    examples:
      - foo
      - bar
      - baz
```

### Affected area/feature

ESC

PR #582, opened 2025-09-01
Add support for fn::concat to concatenate arrays

tehsis:
This PR adds support for a new built-in function `fn::concat` which can be used to concatenate arrays.

```yaml
values:
  foo:
    - 1
    - 2
  bar:
    - 3
    - 4
  baz:
    fn::concat:
        - ${foo}
        - ${bar}
```

results in:

```json
{
  "bar": [
    3,
    4
  ],
  "baz": [
    1,
    2,
    3,
    4
  ],
  "foo": [
    1,
    2
  ]
}
```
Solves #287

### DIFF
--- a/eval/eval.go
+++ b/eval/eval.go
@@ -276,6 +276,7 @@ type exprNode interface {
 // - {Null, Boolean, Number, String}Expr -> literalExpr
 // - InterpolateExpr                     -> interpolateExpr
 // - SymbolExpr                          -> symbolExpr
+// - ConcatExpr                          -> concatExpr
 // - FromBase64Expr                      -> fromBase64Expr
 // - FromJSONExpr                        -> fromJSONExpr
 // - JoinExpr                            -> joinExpr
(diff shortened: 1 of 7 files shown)


eval/eval.go
```diff
@@ -1218,6 +1227,31 @@ func asObjectOrNil(v any) map[string]esc.Value {
...
+
+	v.combine(arrays)
+	if !v.unknown {
+		var result []*value
+		for _, arrayVal := range arrays.repr.([]*value) {
+			if arrayVal.repr != nil {
+				if arr, isArray := arrayVal.repr.([]*value); isArray {
+					result = append(result, arr...)
+				}
+			}
```
pgavlin, line 1248: I _think_ you should be able to just do `append` here w/o the nil + type checks:

```suggestion
			result = append(result, arrayVal.repr.([]*value)...)
```

The type checker should have issued errors otherwise.

eval/testdata/eval/builtin-concat/env.yaml
```diff
@@ -0,0 +1,32 @@
...
+  mixed:
+    fn::concat: [["string", 42], [true, null], [{"key": "value"}]]
+  
+  # Nested arrays (arrays of arrays)
+  nested:
+    fn::concat: [[[1, 2]], [[3, 4]]]
+  
+  # Empty input
+  emptyInput:
+    fn::concat: []
```
pgavlin, line 32: would be good to have a test cases that uses references/symbols, e.g.

```
refs:
  one: [1, 2]
  two: [3, 4]
  concat:
    fn::concat: [${one}, ${two}]
```

pgavlin (approved):
LGTM; just one nit for an extra test case

PR #592, opened 2025-10-02
Add support for Open Approvals

tehsis:
This PR adds support for the new Open Approvals feature in Pulumi ESC.

### DIFF
--- a/cmd/esc/cli/env_open_request.go
+++ b/cmd/esc/cli/env_open_request.go
@@ -22,7 +22,8 @@ func newEnvOpenRequestCmd(envcmd *envCommand) *cobra.Command {
 			"\n" +
 			"This command creates a request to open a protected environment. The request must be\n" +
 			"approved before the environment can be accessed.\n",
-		Hidden: true,
+		Hidden:       true,
+		SilenceUsage: true,
 		RunE: func(cmd *cobra.Command, args []string) error {
 			ctx := context.Background()
 


cmd/esc/cli/env_open_request.go
```diff
@@ -0,0 +1,63 @@
...
+
+	cmd := &cobra.Command{
+		Use:   "open-request [<org-name>/][<project-name>/]<environment-name>[@<version>]",
+		Args:  cobra.ExactArgs(1),
+		Short: "Create a request for opening a protected environment.",
+		Long: "Create a request for opening a protected environment with the given name.\n" +
+			"\n" +
+			"This command creates a request to open a protected environment. The request must be\n" +
+			"approved before the environment can be accessed.\n",
+		SilenceUsage: true,
```
nyobe, line 24: why SilenceUsage?

also, should we set `Hidden: true` to hide the command while the feature is pre-release?

cmd/esc/cli/env_open_request.go
```diff
@@ -0,0 +1,63 @@
...
+				ref.projectName,
+				ref.envName,
+				grantExpirationSeconds,
+				accessDurationSeconds,
+			)
+			if err != nil {
+				return err
+			}
+
+			fmt.Fprintf(envcmd.esc.stdout, "Created environment open request with ID: %s\n", resp.ChangeRequests[0].ChangeRequestID)
```
nyobe, line 49: note to self: there the console should have a route for viewing an open request so that this can output a link similar to https://github.com/pulumi/esc/blob/main/cmd/esc/cli/env.go#L429

nyobe (approved):

cmd/esc/cli/env_open_request.go
nyobe: I added deep linking here: https://github.com/pulumi/pulumi-service/pull/33218

PR #601, opened 2025-11-17
fix rotation patches that contain secret objects

nyobe:
Problem: if a rotator returns a non-primitive secret, ApplyValuePatches fails to produce a valid env definition because it naively wraps secret values with `fn::secret`, and `fn::secret` requires its argument to be a string literal.

For example, if a rotator returns `esc.NewSecret(map[string]esc.Value{"foo": "bar"})` the applied patch will result in something like:
```
values:
  example:
    fn::rotate::test:
      inputs: {}
      state:
        fn::secret:
          foo: bar  # validation error: secret values must be string literals
```

Solution: encode non-primitive secrets as json, and wrap them with `fn::fromJSON` so they get decoded into the expected secret object during evaluation:
```
values:
  example:
    fn::rotate::test:
      inputs: {}
      state:
        fn::fromJSON:
          fn::secret: '{"foo": "bar"}'
```

This came up during https://github.com/pulumi/pulumi-service/pull/34661

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -1,13 +1,8 @@
 ### Improvements
 
-- Added deletion protection for environments:
-  - Use `esc env settings set [<org-name>/][<project-name>/]<environment-name> deletion-protected true` to enable deletion protection
-  - Use `esc env settings get [<org-name>/][<project-name>/]<environment-name> [deletion-protected]` to check the current status
-  - When enabled, environments cannot be deleted until protection is disabled
-  - Deletion protection is disabled by default for new environments
+- Added support for `fn::split` built-in function to split strings into arrays.
+  [#281](https://github.com/pulumi/esc/issues/281)
 
 ### Bug Fixes
-
-- Environment declarations are now returned even in the face of syntax errors.
 
 ### Breaking changes
--- a/ast/expr.go
+++ b/ast/expr.go
@@ -574,6 +574,31 @@ func Concat(arrays *ArrayExpr) *ConcatExpr {
 	}
 }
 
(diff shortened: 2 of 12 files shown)


eval/patch_test.go
```diff
@@ -15,25 +15,84 @@
...
+							"rotatedAt": esc.NewValue("2024-01-01"),
+						}),
+					}),
+				}),
+			},
+		}
+
+		result, err := ApplyValuePatches(source, patches)
+		assert.NoError(t, err)
+		t.Logf("Result YAML:\n%s", string(result))
```
seanyeh: do we want to keep this log?

eval/patch_test.go
```diff
@@ -15,25 +15,84 @@
...
+		// When we have a secret object, fn::secret can't handle it directly because it only accepts string literals.
+		// Therefore we JSON-encode the object and use fn::fromJSON + fn::secret to decode it.
+		actual, err := valueToSecretJSON(esc.NewSecret(map[string]esc.Value{
+			"username": esc.NewValue("admin"),
+			"password": esc.NewValue("secret123"),
+		}))
+		require.NoError(t, err)
+		expected := map[string]any{
+			"fn::fromJSON": map[string]any{
+				"fn::secret": `{"password":"secret123","username":"admin"}`,
```
seanyeh: i know that go randomizes the order of keys, but maybe the json encoder is deterministic? want to make sure this doesn't fail randomly

eval/patch_test.go
nyobe: t.Log output is only displayed by `go test` if the test fails.

I'd keep it to make it easier to tell what went wrong with the patched result if Check fails

PR #625, opened 2026-03-20
Environment level schema validation

seanyeh:
Adds a top-level `schema` key to environments for schema validation

Example
```
schema:
  type: object
  properties:
    foo:
      type: string
  required:
    - foo
values:
  foo: bar
```


seanyeh:
@pgavlin do you have thoughts on a top-level `schema` key for environment level schema validation?

PR #631, opened 2026-04-01
Surface warnings when editing with the CLI

### LINKED ISSUE #626: Surface warnings when editing with the CLI
Warnings are not surfaced when editing environments with the CLI with `esc env edit`

Example:
* `esc env edit <org>/<proj>/<env>`
*  Save environment
```
values:
  x: 1
values: # duplicate top-level values should be a warning
  y: 1
```

fnune:
## Summary

- Add `Severity` field to `EnvironmentDiagnostic` to distinguish warnings from errors
- Display warnings to stderr when editing environments, while still allowing the update to succeed
- Backward compatible: missing/empty severity is treated as an error

image

## Test plan

- Added `env-edit-file-warnings.yaml` test case
- Verified warnings go to stderr, not stdout
- Verified exit code is 0 when only warnings are present

Fixes #626

### DIFF
--- a/cmd/esc/cli/client/apitype.go
+++ b/cmd/esc/cli/client/apitype.go
@@ -37,7 +37,8 @@ type EnvironmentDiagnostic struct {
 }
 
 func (d EnvironmentDiagnostic) IsError() bool {
-	return d.Severity != DiagWarning
+	// Empty severity means error for backward compatibility.
+	return d.Severity == DiagError || d.Severity == ""
 }
 
 func DiagnosticsHaveErrors(diags []EnvironmentDiagnostic) bool {
--- a/cmd/esc/cli/client/apitype.go
+++ b/cmd/esc/cli/client/apitype.go
@@ -22,10 +22,31 @@ import (
 	"github.com/pulumi/esc"
 )
 
+type EnvironmentDiagnosticSeverity string
+
+const (
+	DiagError   EnvironmentDiagnosticSeverity = "error"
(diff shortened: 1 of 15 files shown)


cmd/esc/cli/client/apitype.go
```diff
@@ -22,10 +22,31 @@ import (
...
-	Summary string     `json:"summary,omitempty"`
-	Detail  string     `json:"detail,omitempty"`
+	Range    *esc.Range                    `json:"range,omitempty"`
+	Summary  string                        `json:"summary,omitempty"`
+	Detail   string                        `json:"detail,omitempty"`
+	Severity EnvironmentDiagnosticSeverity `json:"severity,omitempty"`
+}
+
+func (d EnvironmentDiagnostic) IsError() bool {
+	return d.Severity != DiagWarning
```
seanyeh: i would check if severity == error

cmd/esc/cli/client/apitype.go
```diff
@@ -22,10 +22,31 @@ import (
 	"github.com/pulumi/esc"
 )
 
+type EnvironmentDiagnosticSeverity string
```
seanyeh, line 25: is there a reason we use string instead of hcl.DiagnosticSeverity? (I'm fine with either)

seanyeh (approved):

cmd/esc/cli/client/apitype.go
fnune: I added the `== error` check but we also need `== ""` for backward compat.

PR #638, opened 2026-04-21
New built-in functions for late binding interpolation (fn::template and fn::eval)

tehsis:
## Summary

This PR introduces new built-in function to enable late binding interpolation

The `fn::template` built-in function defines a template expression whose interpolations are not resolved immediately. Instead, the template is evaluated later when a consumer environment references it with [`fn::eval`](/docs/esc/environments/syntax/builtin-functions/fn-eval/). This enables late binding — defining reusable expressions in one environment and evaluating them with different values in another.

## Test changes

- [] late-binding unit tests were added

## Risk

None.

## Rollback

Revert the commit

### DIFF
--- a/ast/expr.go
+++ b/ast/expr.go
@@ -728,6 +728,8 @@ type EvalExpr struct {
 	builtinNode
 
 	//TemplateValue Expr
+	// either a symbol or a template expr?
+	// or just allow anything but behave as pass-thru for non-templates?
 }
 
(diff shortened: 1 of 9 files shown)


pgavlin:
this really needs a design doc. do we have one?

tehsis:
> this really needs a design doc. do we have one?

not really, this was meant as a conversation starter rather than an actual PR. Just moved to _draft_

PR #640, opened 2026-04-22
add request ID to error messages in the api client

tgummerer:
If we get a 500 from Pulumi Cloud, we currently get no further indication of what happened, other than "Internal Server Error".  We have a request ID in the response headers from Pulumi Cloud, so we can make use of that, and add it to the error message.

This can help us track down logs from the Cloud to see what happened.

/xref https://github.com/pulumi/pulumi/issues/22690

(We implemented the same thing in pulumi/pulumi in https://github.com/pulumi/pulumi/pull/20653).


fnune (approved):
I don't know if we already have a test harness for this but it would be nice to see additions if there is one. Or try to have Claude create something new...

tgummerer:
> Or try to have Claude create something new...

Good idea, done :heavy_check_mark:

PR #646, opened 2026-05-13
cli: add `esc env schedule` subcommands

### LINKED ISSUE #23036: The Cloud-Ready CLI: `pulumi env schedule`

borisschlosser:
## Summary

Adds `esc env schedule {list,new,get,edit,history,remove}` that wrap the Pulumi Cloud REST schedule endpoints for environments. All eight `/api/esc/environments/{org}/{project}/{env}/schedules{,/{id}{,/history}}` endpoints documented in the Cloud REST API reference are now covered.

- `esc env schedule list <ref>` (alias `ls`) — list scheduled actions; `--utc` formats timestamps in UTC (otherwise local time). `--count N` caps the number of results client-side (the underlying endpoint does not paginate).
- `esc env schedule new <ref>` — create a secret-rotation schedule. Requires exactly one of `--cron` or `--once`.
- `esc env schedule get <ref> <id>` — show details for a single schedule; `--utc` flag.
- `esc env schedule edit <ref> <id>` (aliases `update`, `modify`) — update timing of an existing schedule. Same `--cron` / `--once` semantics as `new`.
- `esc env schedule history <ref> <id>` — show execution history; `--utc` and `--count` flags.
- `esc env schedule remove <ref> <id>` (aliases `rm`, `delete`) — delete a schedule. Interactive type-`remove`-to-confirm prompt unless `--yes` is passed (also honors `PULUMI_SKIP_CONFIRMATIONS=true`), matching the `esc env rm` pattern.

Today only secret-rotation schedules are creatable via the REST API, so no `--action` flag is exposed.

Command and verb naming follows the [CLI Command Guidelines](https://www.notion.so/CLI-Command-Guidelines-350fdbdf1cce80b89367c16b7b421e3d): `list` / `edit` / `remove` are the canonical verbs; established short forms (`ls`, `rm`) and synonyms (`update`/`modify`, `delete`) are added as aliases.

`pause` / `resume` are intentionally **not** exposed in this PR per review feedback — neither is currently surfaced in the cloud UI for ESC secret rotation, so we'll wait on a product decision before adding them.

### Backend constraints surfaced in the CLI

The backend enforces several restrictions that the CLI now mirrors / documents:

- **One schedule per environment.** Backend returns 400 on a second create. The CLI pre-checks via `ListEnvironmentSchedules` and returns a friendly error citing the existing schedule ID. Also documented in `new --help`.
- **Minimum cron interval is once daily.** Documented in `new` / `edit --help` and on the `--cron` flag; the backend's 400 message propagates verbatim if a user slips past.
- **`--once` must be in the future.** Validated client-side as RFC 3339 before any API call.
- **No `--path` flag.** Although the REST API accepts `secretRotationRequest.environmentPath`, the backend currently forces it to `""` (rotate whole environment), so exposing the flag would lie to users. It can be reintroduced when path-scoped rotation lands server-side.

### Follow-ups

- #651 tracks adding `--output json` across ESC commands (raised in review; repo-wide change, not schedule-specific).

### End-to-end testing

Tested against a local backend (`http://localhost:8080`, `dev-org`) across 41 scenarios:

**Happy path:**
- `new --cron` / `new --once` round-trip cleanly; created schedules are visible via `list` / `ls` / `get`.
- `edit` with `--cron` and with `--once` correctly transitions the schedule timing and updates `nextExecution`. Both `update` and `modify` aliases verified.
- `history` returns empty for a schedule with no executions (200 with empty array) — no client-side error.
- `remove` works with all three verbs (`remove`, `rm`, `delete`).
- `--utc` correctly switches the rendered `Schedule:`, `Next execution:`, and `Last executed:` fields between local time and UTC.

**Confirmation flow:**
- `remove` (no `--yes`) prompts to type `remove`; typing anything else exits with `confirmation declined` and the schedule is preserved.
- `remove --yes` skips the prompt.
- `PULUMI_SKIP_CONFIRMATIONS=true` also skips the prompt (matches `esc env rm` behavior).

**Client-side validation rejects before any API call:**
- missing or duplicate `--cron`/`--once` on `new` and `edit`
- malformed `--once` value
- `--once` in the past
- negative `--count` on `list` / `history`
- version refs (`<env>@<rev>`) rejected on all six subcommands with a clear "the X command does not accept versions" error

**Backend errors surface cleanly via `[<code>] <message>`:**
- sub-daily cron (`*/30 * * * *`) → `[400] Bad Request: minimum rotation interval is once daily`
- unknown schedule ID on `get` / `edit` / `remove` / `history` → `[404] Not Found: ScheduledAction '<id>' not found`
- non-existent env → `[404] Not Found`
- second `new` on an env with an existing schedule → friendly pre-check error citing the existing schedule ID

**Observed (non-blocking):**
- The backend allows creating a schedule against an environment with no rotator — a backend-side decision, not a CLI concern.
- `list` on a non-existent env returns a bare `[404] Not Found` (no message body); cosmetic, could be improved server-side.

## Test plan

- [x] 41 manual end-to-end scenarios against a live local backend (see above).
- [x] `make test` passes (includes `testdata/env-schedule-ok.yaml` golden fixture exercising every subcommand and the `ls` alias).
- [x] `make lint` clean (`pulumictl copyright` + `golangci-lint`).
- [x] `make format` clean.

Closes pulumi/pulumi#23036.


cmd/esc/cli/client/apitype.go
```diff
@@ -271,3 +271,40 @@ type EnvironmentSettings struct {
...
+type ScheduledAction struct {
+	ID            string         `json:"id"`
+	OrgID         string         `json:"orgID"`
+	Kind          string         `json:"kind"`
+	Paused        bool           `json:"paused"`
+	Created       string         `json:"created"`
+	Modified      string         `json:"modified"`
+	LastExecuted  string         `json:"lastExecuted"`
+	NextExecution string         `json:"nextExecution"`
+	Definition    map[string]any `json:"definition"`
```
julienp: I think this should use `json.RawMessage`

cmd/esc/cli/env_schedule.go
```diff
@@ -0,0 +1,30 @@
...
+			"A scheduled action runs against an environment on a cron schedule or at a single\n" +
+			"point in time. Today the CLI exposes secret-rotation schedules.\n" +
+			"\n" +
+			"Subcommands exist for listing, creating, pausing, resuming, and removing schedules.",
+		Args:         cobra.NoArgs,
+		SilenceUsage: true,
+	}
+
+	cmd.AddCommand(newEnvScheduleListCmd(env))
+	cmd.AddCommand(newEnvScheduleNewCmd(env))
```
julienp, line 25: We don't have an `edit` command here, did we miss that when listing all the commands to add or is this intentional?

We can always follow up at a later date with `edit` though, since you can delete + create.

cmd/esc/cli/client/apitype.go
```diff
@@ -271,3 +271,40 @@ type EnvironmentSettings struct {
 type PatchEnvironmentSettingsRequest struct {
 	DeletionProtected *bool `json:"deletionProtected,omitempty"`
 }
+
+// ScheduledAction describes a scheduled action attached to an environment.
+//
+// Time fields (Created, Modified, LastExecuted, NextExecution, ScheduleOnce) are kept as
+// raw ISO 8601 strings on the wire; callers can parse them as needed.
+type ScheduledAction struct {
```
julienp, line 280: [nonblocking] Question for the larger esc team for later: We will have this defined in the pulumi CLI sdk apitypes once everything is merged over there, with this exact shape (except for the `json.RawMessage` flagged below). Maybe we should come back later to import this here.

cmd/esc/cli/env_schedule_pause.go
```diff
@@ -0,0 +1,54 @@
...
+
+import (
+	"context"
+	"errors"
+	"fmt"
+
+	"github.com/spf13/cobra"
+)
+
+func newEnvSchedulePauseCmd(env *envCommand) *cobra.Command {
```
julienp: For deployment schedules we decided to not expose `pause` because this is not surfaced in the cloud UI for deployments.

Looking at the cloud UI for ESC secret rotation (I think that's where these schedules are used), I don't see a way of pausing them, so I think we should remove this command for now and wait for a production decision on this. Same for `resume`.

I might have simply missed the functionality in the UI though ...

cmd/esc/cli/env_schedule_rm.go
```diff
@@ -0,0 +1,53 @@
...
+	"context"
+	"errors"
+	"fmt"
+
+	"github.com/spf13/cobra"
+)
+
+func newEnvScheduleRmCmd(env *envCommand) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "rm [<org-name>/][<project-name>/]<environment-name> <schedule-id>",
```
julienp: We've made this require confirmation either via a `--yes` flag or an interactive prompt that requires the user to type in `remove` to confirm. See https://github.com/pulumi/pulumi/pull/23175 for an example.

cmd/esc/cli/env_schedule_list.go
```diff
@@ -0,0 +1,115 @@
...
+)
+
+func newEnvScheduleListCmd(env *envCommand) *cobra.Command {
+	var (
+		utc   bool
+		count int
+	)
+
+	cmd := &cobra.Command{
+		Use:     "list [<org-name>/][<project-name>/]<environment-name>",
```
julienp, line 23: It is a little weird to have this be a `list` command when there can only be exactly one schedule for an env. Maybe this should be `get` instead?

cmd/esc/cli/env_schedule_list.go
```diff
@@ -0,0 +1,115 @@
...
+				resp.Schedules = resp.Schedules[:count]
+			}
+
+			printSchedules(env.esc.stdout, resp, utcFlag(utc))
+			return nil
+		},
+	}
+
+	cmd.Flags().BoolVar(&utc, "utc", false, "display times in UTC")
+	cmd.Flags().IntVar(&count, "count", 0, "the maximum number of schedules to return (all if unset)")
```
julienp, line 64: I would probably remove this since we have at most 1 schedule and we print it in a "inline" way, ie not a list in table format

cmd/esc/cli/env_schedule_list.go
```diff
@@ -0,0 +1,115 @@
...
+	"fmt"
+	"io"
+	"time"
+
+	"github.com/spf13/cobra"
+
+	"github.com/pulumi/esc/cmd/esc/cli/client"
+)
+
+func newEnvScheduleListCmd(env *envCommand) *cobra.Command {
```
julienp, line 16: Most pulumi CLI commands have an `--output` flag that defaults to text output and can make us print json with `--output json`.

Should ESC commands have this too? That might be a separate piece of work though, since other pre-existing commands don't seem to have any json output either.

cmd/esc/cli/env_schedule_remove.go
```diff
@@ -0,0 +1,53 @@
...
+import (
+	"context"
+	"errors"
+	"fmt"
+
+	"github.com/spf13/cobra"
+)
+
+func newEnvScheduleRmCmd(env *envCommand) *cobra.Command {
+	cmd := &cobra.Command{
```
julienp, line 19: should be `remove` with `rm` and `delete` as aliases

cmd/esc/cli/env_schedule_list.go
```diff
@@ -0,0 +1,115 @@
...
+func printSchedule(stdout io.Writer, s client.ScheduledAction, utc utcFlag) {
+	fmt.Fprintf(stdout, "ID: %s\n", s.ID)
+	fmt.Fprintf(stdout, "Kind: %s\n", s.Kind)
+	schedule := s.ScheduleCron
+	if schedule == "" {
+		schedule = formatScheduleTime(s.ScheduleOnce, utc)
+	}
+	if schedule == "" || schedule == "never" {
+		schedule = "<unknown>"
+	}
```
julienp: [nonblocking] I don't love checking for `never` below, maybe something like would be more readable:

```
var schedule string
switch {
case s.ScheduleCron != "":
    schedule = s.ScheduleCron
case s.ScheduleOnce != "":
    schedule = formatScheduleTime(s.ScheduleOnce, utc)
default:
    schedule = "<unknown>"
}
```

cmd/esc/cli/env_schedule_new.go
```diff
@@ -0,0 +1,97 @@
...
+				t, err := time.Parse(time.RFC3339, once)
+				if err != nil {
+					return fmt.Errorf("--once must be an ISO 8601 / RFC 3339 timestamp: %w", err)
+				}
+				if !t.After(time.Now()) {
+					return errors.New("--once must be a timestamp in the future")
+				}
+			}
+
+			existing, err := env.esc.client.ListEnvironmentSchedules(ctx, ref.orgName, ref.projectName, ref.envName)
```
julienp, line 67: [nonblocking] Is the limit of 1 schedule server imposed? If so, we can probably drop this check, but maybe this allows us to print a clearer error message below?

cmd/esc/cli/client/apitype.go
borisschlosser: Done — switched `Definition` to `json.RawMessage`.

cmd/esc/cli/env_schedule.go
borisschlosser: Added `esc env schedule edit` (with `update` / `modify` aliases) in a follow-up commit — it wraps the PATCH endpoint.

cmd/esc/cli/client/apitype.go
borisschlosser: Ack — agreed, this is a candidate for cross-repo consolidation once the pulumi CLI SDK apitypes land.

cmd/esc/cli/env_schedule_pause.go
borisschlosser: Removed `pause` and `resume` (commands and client methods). We can add them back once there's a product decision and the UI exposes the same.

cmd/esc/cli/env_schedule_rm.go
borisschlosser: Done — `remove` now prompts the user to type `remove` to confirm and accepts `--yes` / `PULUMI_SKIP_CONFIRMATIONS` to skip. Pattern matches `esc env rm`.

cmd/esc/cli/env_schedule_list.go
borisschlosser: Kept both `list` and `get <id>` to mirror the REST API endpoints (List and Read). Even if the backend enforces 1-per-env today, the CLI shouldn't bake that constraint in — and the CLI Command Guidelines treat them as separate verbs.

cmd/esc/cli/env_schedule_list.go
borisschlosser: Keeping `--count` for now — symmetric with `history` which clearly benefits from it, and harmless on `list` while we're constrained to 1-per-env.

cmd/esc/cli/env_schedule_list.go
borisschlosser: Filed #651 to track `--output json` support across ESC commands — agreed it's a repo-wide change rather than schedule-specific.

cmd/esc/cli/env_schedule_remove.go
borisschlosser: Done — renamed to `remove` with `rm` and `delete` as aliases.

cmd/esc/cli/env_schedule_list.go
borisschlosser: Applied your suggested switch.

cmd/esc/cli/env_schedule_new.go
borisschlosser: Keeping the pre-check — the friendlier error citing the existing schedule ID is exactly the reason. Without it the user just sees the backend's generic 400.

cmd/esc/cli/env_schedule_list.go
julienp: Sounds good!

julienp (approved):

PR #647, opened 2026-05-13
cli: add `esc env webhook` subcommands

### LINKED ISSUE #23030: The Cloud-Ready CLI: `pulumi env webhook`

borisschlosser:
### Description

Adds CLI commands to manage environment webhooks via the Pulumi Cloud REST API. All commands are marked `[EXPERIMENTAL]` in their long descriptions.

- `esc env webhook list <env>` (alias `ls`) — list webhooks attached to an environment. `--count N` caps results client-side.
- `esc env webhook get <env> <name>` — print a single webhook (including the `Event groups:` line).
- `esc env webhook new <env> <display-name>` — create a webhook. The positional is the human-readable display name; the service generates the unique identifier, which is printed on success and is the name to pass to the other subcommands. Required: `--url`. Optional: `--format` (default `raw`), `--event` (repeatable), `--group` (repeatable), `--active` (default true), `--secret`.
- `esc env webhook edit <env> <name>` — update a webhook. The CLI fetches the current state, applies supplied flag overrides on top of it, and submits the merged result (the service's PATCH endpoint is effectively a PUT). `--event` replaces the event list; `--add-event` / `--remove-event` merge against the current value and cannot be combined with `--event`. The same triplet exists for groups: `--group`, `--add-group`, `--remove-group`. `--remove-secret` clears a stored secret (mutually exclusive with `--secret`; an empty `--secret` is treated as "leave unchanged" by the service).
- `esc env webhook rm <env> <name>` — delete a webhook.
- `esc env webhook ping <env> <name>` — trigger a synthetic delivery and print the result.
- `esc env webhook delivery list <env> <name>` (alias `ls`) — list deliveries recorded for a webhook (with `--utc`, `--count N`).

List output uses `cmdutil.Table` from `pulumi/pulumi`'s `sdk/v3/go/common/util/cmdutil` for consistency with `pulumi/pulumi`. All commands accept the esc-native positional environment ref (`[org/][project/]env`) and reject `@version`.

List-command naming follows the [CLI Command Guidelines](https://www.notion.so/CLI-Command-Guidelines-350fdbdf1cce80b89367c16b7b421e3d): `list` is the canonical verb, `ls` is an alias. The underlying endpoints do not paginate, so `--count` is applied client-side.

### Naming

The user-facing terminology for "events a webhook subscribes to" is `event` / `event groups`, matching the web UI. The CLI flags are `--event` / `--group` (plus the `--add-*` / `--remove-*` variants on `edit`) and the displayed lines are `Events` / `Event groups`. The wire format still uses the existing `filters` / `groups` JSON keys.

Per the Slack agreement on webhook naming, the API `name` field (the unique identifier) is always sent as `""` on create so the service generates it; `displayName` is required and arrives via the positional argument on `new`.

### Client-side validation

`new` and `edit` validate inputs before calling the service to give faster, clearer errors:

- **`--format`** — must be one of `raw`, `slack`, `ms_teams`, `pulumi_deployments`.
- **`--url`** — coupled to the format that will be in effect after the edit:
  - `raw`, `ms_teams`: any `http(s)` URL with a host.
  - `slack`: must begin with `https://hooks.slack.com/`.
  - `pulumi_deployments`: must be of the form `<project>/<stack>`.
- Event and group values are intentionally **not** validated client-side — the supported set evolves frequently, so we rely on the service for that.

### Related

Closes pulumi/pulumi#23030.

Follow-ups raised during review:

- #651 — repo-wide `--output json` support across `esc` commands.
- #652 — consolidate `SilenceUsage` / `SilenceErrors` on the esc root command (matches `pulumi/pulumi`).
- #653 — migrate remaining `esc` list commands to `cmdutil.Table`.

### End-to-end testing

Verified against a local `pulumi-service` at `http://localhost:8080`, org `dev-org`, env `dev-org/webhook-test/env`. A small Go HTTP sink on `127.0.0.1:9117` received deliveries (reached from the service via `127.0.0.1.nip.io`, which resolves to 127.0.0.1 and satisfies the service's payload-URL regex).

| Command | Cases | Result |
| --- | --- | --- |
| `webhook new` | happy path, `--group`, event+group combo, slack URL rejection, missing `--url`, invalid group (server-side), empty display name | ✅ |
| `webhook list` / `ls` | full listing (table renders via `cmdutil.Table`), `--count` limit, negative `--count` rejected | ✅ |
| `webhook get` | populated + empty `Event groups`, missing webhook → 404 | ✅ |
| `webhook edit` | `--active`, `--display-name`, `--add-event` / `--remove-event` merge, `--event` replace, event mutual exclusion, `--add-group` / `--remove-group`, group mutual exclusion, `--secret` / `--remove-secret`, secret mutual exclusion, `--format` + `--url` swap to `slack` and back | ✅ |
| `webhook ping` | sink received POST with `Pulumi-Webhook-Kind: ping`; CLI printed delivery summary | ✅ |
| `webhook delivery list` / `ls` | ping + real `environment_revision` delivery shown; `--utc` converts timestamps correctly | ✅ |
| `webhook rm` | removed all webhooks, list emptied | ✅ |
| **Real event delivery** | editing the env fired `environment_revision` deliveries to every registered webhook | ✅ |

Notable findings (none are CLI bugs):

1. The service's `cleanupSubscriptions` strips individual filters when a group that already covers them is added (e.g. adding the `environments` group removes the `environment_revision_*` filters). The CLI reports the deduped state correctly.
2. `webhook rm` of a non-existent webhook returns exit 0 — the service's DELETE handler is idempotent (204 even when the hook is missing). The CLI faithfully reports what the service did. Consistent with REST DELETE semantics.
3. The service's payload-URL regex requires a TLD; bare `localhost` is rejected. Use `127.0.0.1.nip.io` or any host with a dotted TLD when testing locally.

### Unit tests

- Golden fixtures `cmd/esc/cli/testdata/env-webhook-ok.yaml` and `env-webhook-edit-events.yaml` cover the happy path (including both `list` and the `ls` alias and the new `Event groups:` line) and `--add-event` / `--remove-event` merging.
- Unit tests `cmd/esc/cli/env_webhook_validate_test.go` cover the format / URL validators.
- `make test`, `make lint` pass.

### DIFF
--- a/cmd/esc/cli/env_webhook_delivery.go
+++ b/cmd/esc/cli/env_webhook_delivery.go
@@ -18,7 +18,7 @@ func newEnvWebhookDeliveryCmd(env *envCommand) *cobra.Command {
 		SilenceUsage: true,
 	}
 
-	cmd.AddCommand(newEnvWebhookDeliveryLsCmd(env))
+	cmd.AddCommand(newEnvWebhookDeliveryListCmd(env))
 
 	return cmd
 }
--- a/cmd/esc/cli/env_webhook_delivery.go
+++ b/cmd/esc/cli/env_webhook_delivery.go
@@ -1,0 +1,24 @@
+// Copyright 2026, Pulumi Corporation.
+
+package cli
+
+import (
+	"github.com/spf13/cobra"
+)
+
+func newEnvWebhookDeliveryCmd(env *envCommand) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "delivery",
+		Short: "Manage environment webhook deliveries",
+		Long: "Manage environment webhook deliveries\n" +
+			"\n" +
+			"A delivery is a single attempt to deliver a webhook payload. Subcommands exist for\n" +
+			"listing the deliveries recorded for a given webhook.",
+		Args:         cobra.NoArgs,
+		SilenceUsage: true,
+	}
+
+	cmd.AddCommand(newEnvWebhookDeliveryLsCmd(env))
+
+	return cmd
+}
--- a/cmd/esc/cli/env_webhook_delivery.go
+++ b/cmd/esc/cli/env_webhook_delivery.go
@@ -10,10 +10,9 @@ func newEnvWebhookDeliveryCmd(env *envCommand) *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   "delivery",
 		Short: "Manage environment webhook deliveries",
-		Long: "Manage environment webhook deliveries\n" +
+		Long: "[EXPERIMENTAL] Manage environment webhook deliveries\n" +
 			"\n" +
-			"A delivery is a single attempt to deliver a webhook payload. Subcommands exist for\n" +
-			"listing the deliveries recorded for a given webhook.",
+			"A delivery is a single attempt to deliver a webhook payload.",
 		Args:         cobra.NoArgs,
 		SilenceUsage: true,
 	}
--- a/cmd/esc/cli/env_webhook_delivery_list.go
+++ b/cmd/esc/cli/env_webhook_delivery_list.go
@@ -5,12 +5,13 @@ package cli
 import (
 	"context"
 	"errors"
-	"fmt"
 	"io"
-	"text/tabwriter"
+	"strconv"
 	"time"
 
 	"github.com/spf13/cobra"
+
+	"github.com/pulumi/pulumi/sdk/v3/go/common/util/cmdutil"
 
 	"github.com/pulumi/esc/cmd/esc/cli/client"
 )
@@ -68,8 +69,8 @@ func newEnvWebhookDeliveryListCmd(env *envCommand) *cobra.Command {
 		},
 	}
 
-	cmd.Flags().BoolVar(&utc, "utc", false, "display times in UTC")
-	cmd.Flags().IntVar(&count, "count", 0, "the maximum number of deliveries to return (all if unset)")
+	cmd.Flags().BoolVar(&utc, "utc", false, "Display times in UTC")
+	cmd.Flags().IntVar(&count, "count", 0, "The maximum number of deliveries to return (all if unset)")
 
 	return cmd
 }
@@ -78,12 +79,21 @@ func printWebhookDeliveries(stdout io.Writer, ds []client.EnvironmentWebhookDeli
 	if len(ds) == 0 {
 		return
 	}
-	w := tabwriter.NewWriter(stdout, 0, 0, 2, ' ', 0)
-	fmt.Fprintln(w, "ID\tKIND\tTIMESTAMP\tRESPONSE\tDURATION (ms)")
+	rows := make([]cmdutil.TableRow, 0, len(ds))
 	for _, d := range ds {
 		ts := time.Unix(d.Timestamp, 0)
-		fmt.Fprintf(w, "%s\t%s\t%s\t%d\t%d\n",
-			d.ID, d.Kind, utc.time(ts).Format(time.RFC3339), d.ResponseCode, d.Duration)
+		rows = append(rows, cmdutil.TableRow{
+			Columns: []string{
+				d.ID,
+				d.Kind,
+				utc.time(ts).Format(time.RFC3339),
+				strconv.FormatInt(d.ResponseCode, 10),
+				strconv.FormatInt(d.Duration, 10),
+			},
+		})
 	}
-	_ = w.Flush()
+	_ = cmdutil.FprintTable(stdout, cmdutil.Table{
+		Headers: []string{"ID", "KIND", "TIMESTAMP", "RESPONSE", "DURATION (ms)"},
+		Rows:    rows,
+	})
 }
--- a/cmd/esc/cli/env_webhook_delivery_list.go
+++ b/cmd/esc/cli/env_webhook_delivery_list.go
@@ -25,7 +25,7 @@ func newEnvWebhookDeliveryListCmd(env *envCommand) *cobra.Command {
 		Use:     "list [<org-name>/][<project-name>/]<environment-name> <webhook-name>",
 		Aliases: []string{"ls"},
 		Short:   "List environment webhook deliveries.",
-		Long: "List environment webhook deliveries\n" +
+		Long: "[EXPERIMENTAL] List environment webhook deliveries\n" +
 			"\n" +
 			"This command lists the deliveries recorded for the named webhook.\n",
 		Args:         cobra.ExactArgs(2),
--- a/cmd/esc/cli/env_webhook_delivery_list.go
+++ b/cmd/esc/cli/env_webhook_delivery_list.go
@@ -15,12 +15,16 @@ import (
 	"github.com/pulumi/esc/cmd/esc/cli/client"
 )
 
-func newEnvWebhookDeliveryLsCmd(env *envCommand) *cobra.Command {
-	var utc bool
+func newEnvWebhookDeliveryListCmd(env *envCommand) *cobra.Command {
+	var (
+		utc   bool
+		count int
+	)
 
 	cmd := &cobra.Command{
-		Use:   "ls [<org-name>/][<project-name>/]<environment-name> <webhook-name>",
-		Short: "List environment webhook deliveries.",
+		Use:     "list [<org-name>/][<project-name>/]<environment-name> <webhook-name>",
+		Aliases: []string{"ls"},
+		Short:   "List environment webhook deliveries.",
 		Long: "List environment webhook deliveries\n" +
 			"\n" +
 			"This command lists the deliveries recorded for the named webhook.\n",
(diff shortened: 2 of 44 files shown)


cmd/esc/cli/env_webhook_rm.go
```diff
@@ -0,0 +1,54 @@
...
+func newEnvWebhookRmCmd(env *envCommand) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "rm [<org-name>/][<project-name>/]<environment-name> <webhook-name>",
+		Short: "Remove an environment webhook.",
+		Long: "Remove an environment webhook\n" +
+			"\n" +
+			"This command removes the named webhook from the environment.\n",
+		Args:         cobra.ExactArgs(2),
+		SilenceUsage: true,
+		RunE: func(cmd *cobra.Command, args []string) error {
```
julienp: We should ask for confirmation with `--yes` flag or retyping the name in interactive mode, see https://github.com/pulumi/pulumi/pull/23102

cmd/esc/cli/env_webhook_delivery.go
```diff
@@ -0,0 +1,24 @@
...
+func newEnvWebhookDeliveryCmd(env *envCommand) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "delivery",
+		Short: "Manage environment webhook deliveries",
+		Long: "Manage environment webhook deliveries\n" +
+			"\n" +
+			"A delivery is a single attempt to deliver a webhook payload. Subcommands exist for\n" +
+			"listing the deliveries recorded for a given webhook.",
+		Args:         cobra.NoArgs,
+		SilenceUsage: true,
```
tgummerer, line 17: Why do we have `SilenceUsage: true` everywhere? That seems useful when running commands (and for consistency with `pulumi/pulumi` we shouldn't have this set unless there's a very good reason for it)

cmd/esc/cli/env_webhook_delivery_list.go
```diff
@@ -0,0 +1,89 @@
...
+)
+
+func newEnvWebhookDeliveryListCmd(env *envCommand) *cobra.Command {
+	var (
+		utc   bool
+		count int
+	)
+
+	cmd := &cobra.Command{
+		Use:     "list [<org-name>/][<project-name>/]<environment-name> <webhook-name>",
```
tgummerer, line 26: This is a bit different from the `stack` and `org` webhooks, in that it requires specifying an argument for the environment name, but I think that's fine, as we can't infer environments here.

cmd/esc/cli/env_webhook_edit.go
```diff
@@ -0,0 +1,195 @@
...
+			}
+
+			fmt.Fprintf(env.esc.stdout, "Updated webhook %s for %s/%s/%s\n",
+				w.Name, ref.orgName, ref.projectName, ref.envName)
+			return nil
+		},
+	}
+
+	cmd.Flags().StringVar(&url, "url", "", "the payload URL to deliver events to")
+	cmd.Flags().StringVar(&displayName, "display-name", "", "the display name")
```
tgummerer: I've just called this `--name` everywhere, as that's what it's called in the UI.

cmd/esc/cli/env_webhook_delivery_list.go
```diff
@@ -0,0 +1,89 @@
...
+			if count > 0 && len(deliveries) > count {
+				deliveries = deliveries[:count]
+			}
+
+			printWebhookDeliveries(env.esc.stdout, deliveries, utcFlag(utc))
+			return nil
+		},
+	}
+
+	cmd.Flags().BoolVar(&utc, "utc", false, "display times in UTC")
```
tgummerer: All of the descriptions should start with uppercase letters for consistency.

cmd/esc/cli/env_webhook_delivery_list.go
```diff
@@ -0,0 +1,89 @@
...
+	cmd.Flags().IntVar(&count, "count", 0, "the maximum number of deliveries to return (all if unset)")
+
+	return cmd
+}
+
+func printWebhookDeliveries(stdout io.Writer, ds []client.EnvironmentWebhookDelivery, utc utcFlag) {
+	if len(ds) == 0 {
+		return
+	}
+	w := tabwriter.NewWriter(stdout, 0, 0, 2, ' ', 0)
```
tgummerer: For lists we've been using tables rather than the tabwriter everywhere.

cmd/esc/cli/env_webhook_edit.go
```diff
@@ -0,0 +1,195 @@
...
+	cmd.Flags().StringVar(&format, "format", "", "the payload format")
+	cmd.Flags().StringArrayVar(&events, "event", nil, "replace the subscribed events (repeatable)")
+	cmd.Flags().BoolVar(&active, "active", true, "whether the webhook is active")
+	cmd.Flags().StringVar(&secret, "secret", "", "shared secret used to sign deliveries")
+	cmd.Flags().BoolVar(&removeSecret, "remove-secret", false, "clear the existing shared secret")
+	cmd.Flags().StringArrayVar(&addEvents, "add-event", nil, "subscribe to an additional event (repeatable)")
+	cmd.Flags().StringArrayVar(&removeEvents, "remove-event", nil, "unsubscribe from an event (repeatable)")
```
tgummerer: Doesn't this also need event groups? Or do esc webhooks not have that?

tgummerer (commented):
I've not really reviewed the code itself (I trust it's fine), but mostly the interface.

Something else that stood out to me here is that none of these commands support json output as of now.  Is that planned as a followup?

cmd/esc/cli/env_webhook_delivery.go
borisschlosser: Good catch — `SilenceUsage: true` (and `SilenceErrors: true`) is set only on the root command in `pulumi/pulumi` and inherited via cobra. In esc the convention is the opposite: not set on the root, redundantly set on every subcommand. The webhook commands just match the existing esc pattern.

This is worth fixing repo-wide rather than only for the webhook commands. Filed #652 to track consolidating both flags on the esc root and removing them from subcommands.

cmd/esc/cli/env_webhook_edit.go
borisschlosser: Reworked per the Slack agreement on webhook naming (https://pulumi.slack.com/archives/C0B24LKK8UT/p1778832281907679?thread_ts=1778830622.106859&cid=C0B24LKK8UT): `new` now takes a `<webhook-display-name>` positional and sends `Name: ""` so the service generates the identifier. The success message prints the generated name so users know what to pass to `edit`/`get`/`rm`/`ping`/`delivery list`. `edit` keeps the existing `<webhook-name>` positional and `--display-name` flag.

cmd/esc/cli/env_webhook_delivery_list.go
borisschlosser: Capitalized all flag descriptions across the 9 webhook command files in 8b7ce7d to match the pulumi/pulumi convention.

cmd/esc/cli/env_webhook_delivery_list.go
borisschlosser: Switched `env webhook list` and `env webhook delivery list` to `cmdutil.Table` + `FprintTable` in 8b7ce7d. Filed #653 to migrate the rest of the esc list commands (which currently use a mix of plain printing and key/value blocks) so the repo lands on a single convention.

cmd/esc/cli/env_webhook_edit.go
borisschlosser: Confirmed via service code: env webhooks do support event groups. Added full support in 8b7ce7d:

- `new`: `--group` flag (repeatable).
- `edit`: `--group` / `--add-group` / `--remove-group` (mirroring the existing `--event` flags, with the same "replace vs. merge" exclusion rule).
- `get`: prints an `Event groups:` line.

The service validates names; for env webhooks the allowed groups are `environments` and `change_requests`.

borisschlosser:
@tgummerer on JSON output: that's tracked as a repo-wide follow-up at #651 — we want `--output json` across all esc `get`/`list` commands rather than only the webhook ones, so the webhook commands will pick it up when that lands.

cmd/esc/cli/env_webhook_new.go
```diff
@@ -0,0 +1,108 @@
...
+			return nil
+		},
+	}
+
+	cmd.Flags().StringVar(&url, "url", "", "the payload URL to deliver events to (required)")
+	cmd.Flags().StringVar(&displayName, "display-name", "", "the display name (defaults to the webhook name)")
+	cmd.Flags().StringVar(&format, "format", "raw", "the payload format")
+	cmd.Flags().StringArrayVar(&events, "event", nil, "event types to subscribe to (repeatable)")
+	cmd.Flags().BoolVar(&active, "active", true, "whether the webhook is active")
+	cmd.Flags().StringVar(&secret, "secret", "", "shared secret used to sign deliveries")
```
tgummerer: These are not capitalized.  Please make sure to apply comments to all files.

borisschlosser: All flag descriptions across the 9 `env_webhook*.go` files were capitalized in 8b7ce7d (pushed after this comment landed). `--display-name` specifically went away because that field is now the positional `<webhook-display-name>` on `new`. Quick verification:

```
$ grep -h 'cmd.Flags().*Var.*"[a-z]' cmd/esc/cli/env_webhook*.go
(no output)
```

tgummerer (approved):

PR #649, opened 2026-05-14
cmd/esc: add `env provider <cloud>-login {static,oidc}` subcommands for AWS, Azure, GCP

borisschlosser:
## Summary

Adds an `esc env provider` parent command. Each cloud (`aws-login`, `azure-login`, `gcp-login`) is itself a parent that exposes two subcommands — `static` and `oidc` — for writing the consuming side of a login-provider block into an environment definition.

```
esc env provider aws-login   static <env-ref> <access-key-id> <secret-access-key>           [--session-token]                       [--create] [--path]
esc env provider aws-login   oidc   <env-ref> <role-arn> <session-name>                     [--duration] [--policy-arn ...] [--subject-attribute ...]      [--create] [--path]

esc env provider azure-login static <env-ref> <tenant-id> <subscription-id> <client-id> <client-secret>                                                   [--create] [--path]
esc env provider azure-login oidc   <env-ref> <tenant-id> <subscription-id> <client-id>     [--subject-attribute ...]               [--create] [--path]

esc env provider gcp-login   static <env-ref> <project-number> <access-token>               [--service-account] [--token-lifetime]                       [--create] [--path]
esc env provider gcp-login   oidc   <env-ref> <project-number> --workload-pool-id POOL --provider-id PROV --service-account SA [--region] [--token-lifetime] [--subject-attribute ...] [--create] [--path]
```

Each subcommand fetches the current env definition, merges an `fn::open::<provider>-login` block at `--path` under `values` (defaults `aws.login`, `azure.login`, `gcp.login`), and writes back via the existing `updateEnvironment` flow. Existing blocks at the target path are replaced wholesale.

- Sensitive fields (`secretAccessKey`, `sessionToken`, `clientSecret`, GCP inner `accessToken`) are wrapped in `fn::secret`.
- For Azure, **`clientSecret` is required in `static` mode** (it's the authentication method) and **must be absent in `oidc` mode** (federation supplies the trust path). The two modes are mutually exclusive; the backend rejects environments that have both or neither.
- `--draft[=<change-request-id>]` is supported, mirroring `esc env set`.
- `--create` creates the target environment if it doesn't already exist (no-op when it does).
- Repeatable flags (`--policy-arn`, `--subject-attribute`) accept the value multiple times rather than comma-separated, so ARNs and other values containing commas pass through intact.

Backend schema diagnostics (e.g. invalid `duration`, unknown `subject-attribute`, malformed ARN) are rendered with source-code context via the same hcl-based diagnostic writer used by `esc env edit`, and the command exits non-zero when error-severity diagnostics are returned. Client-side validation of field formats was deliberately not added so the CLI doesn't duplicate (and drift from) the backend schema.

Tracks pulumi/pulumi#23043 (parent epic pulumi/pulumi#22959).

## Missing-environment behavior

Default is fail-fast: if the target env doesn't exist, exit with a clear error pointing at `--create` or `esc env init`. No interactive prompts, no silent auto-create on typos. Agents and CI get deterministic behavior.

```
$ esc env provider aws-login static org/proj/missing AKIA... shhh
Error: environment org/proj/missing does not exist; pass --create to create it, or run `esc env init org/proj/missing` first

$ esc env provider aws-login static org/proj/missing AKIA... shhh --create
Environment created: org/proj/missing
Provider updated.
```

## Cloud-side OIDC resources are out of scope (by design)

OIDC requires resources on the cloud provider side — IAM role + trust policy for AWS, Workload Identity Pool + Provider for GCP, App Registration + federated credentials for Azure. **`esc env provider` writes only the consuming side of that trust relationship**; provisioning the cloud-side resources belongs in a different command. The canonical workflow is:

1. **Provision cloud-side OIDC trust** with Pulumi, an existing IaC pipeline, or a future dedicated CLI command. The output is a stable identifier (role ARN, workload pool ID, app client ID).
2. **Write the ESC side** by passing those identifiers into `esc env provider <cloud>-login oidc`.

### Why these are two commands, not one

1. **Different credential authority.** Provisioning needs cloud-admin/IAM credentials with permission to create roles, federations, and trust policies. Writing the ESC side needs only an ESC API token. Fusing them means the operator running `env provider oidc` has to hold both at the same time — a strictly larger blast radius than either operation needs on its own, and a privilege pattern most agent/CI use cases can't satisfy.
2. **Different ownership boundary.** Cloud-side OIDC trust is provisioned by cloud admins (often under stricter change-control than developer workflows), and is shared 1:N with many ESC environments. ESC config is owned by environment owners and changes per env/per rotation. A single command muddles who owns which side of the change; two commands let each role do its job.
3. **Different lifecycle and idempotency contract.** Cloud resources are created once and rarely updated; re-provisioning them on every ESC config change is wrong. ESC config changes are frequent and idempotent. A combined command would have to decide on every invocation whether to leave cloud resources alone, update them, or recreate them — a hard product problem we shouldn't smuggle into a config-writing command.
4. **Mirrors the platform.** Cloud-side provisioning and ESC-side config are already separate concerns on the platform. Two CLI commands map cleanly onto that separation; one CLI command bridging both makes the CLI a leakier abstraction than the platform underneath it.
5. **Consistency with `static`.** `static` is a pure ESC-API write. If `oidc` quietly performed cloud-side provisioning under the same parent, two siblings would have radically different blast radii and required privileges despite identical surface — a confusing footgun for anyone scanning `--help`.
6. **Composability for agents and pipelines.** Separate commands compose: provision once, write the consumer N times across orgs/envs/paths. A unified command forces every ESC-write to also re-traverse cloud-admin auth, even when no provisioning is needed. Atomic single-purpose commands are also easier for agents to retry safely (an ESC-write retry can't accidentally re-trigger cloud-resource creation).
7. **Failure-mode separation.** ESC write failures are recoverable by retry. Cloud-resource provisioning failures often leave dangling state (half-created role, federation rule pointing at no role) requiring manual cleanup. Bundling them means one transient cloud-API failure leaves the user with a partially-written ESC env *and* partially-provisioned cloud state — strictly worse than the bounded blast radius of either failing alone.

If demand emerges for a single-step \"set up cloud resources and write the ESC side\" UX, it should ship as a separate top-level command that calls both layers in sequence — keeping the per-layer commands available for the cases where they're sufficient.

## YAML shapes produced

```yaml
# aws-login static at --path aws.login
values:
  aws:
    login:
      fn::open::aws-login:
        static:
          accessKeyId: AKIA...
          secretAccessKey:
            fn::secret: ...
          sessionToken:                       # only when --session-token is given
            fn::secret: ...

# aws-login oidc
values:
  aws:
    login:
      fn::open::aws-login:
        oidc:
          roleArn: arn:aws:iam::123:role/r
          sessionName: pulumi-esc
          duration: 1h                        # optional
          policyArns: [arn1, arn2]            # optional
          subjectAttributes: [env, team]      # optional

# azure-login static at --path azure.login  (clientSecret is required)
values:
  azure:
    login:
      fn::open::azure-login:
        clientId: ...
        tenantId: ...
        subscriptionId: ...
        clientSecret:
          fn::secret: ...

# azure-login oidc  (clientSecret intentionally absent — federation supplies trust)
values:
  azure:
    login:
      fn::open::azure-login:
        clientId: ...
        tenantId: ...
        subscriptionId: ...
        oidc: true
        subjectAttributes: [env]              # optional

# gcp-login static at --path gcp.login
values:
  gcp:
    login:
      fn::open::gcp-login:
        project: 123456789
        accessToken:
          accessToken:
            fn::secret: ...
          serviceAccount: sa@proj.iam.gserviceaccount.com   # optional, for impersonation
          tokenLifetime: 1h                                  # optional

# gcp-login oidc
values:
  gcp:
    login:
      fn::open::gcp-login:
        project: 123456789
        oidc:
          workloadPoolId: pool
          providerId: provider
          serviceAccount: sa@proj.iam.gserviceaccount.com
          region: us-central1                 # optional
          tokenLifetime: 1h                   # optional
          subjectAttributes: [env]            # optional
```

## End-to-end test results

Ran against a local backend with a personal access token in org `dev-org`. All envs were namespaced under `esc-provider-e2e/e2e-<timestamp>-*`, created on demand, and removed at the end of the run. No cloud-provider resources were created — credentials/IDs passed to the commands were placeholders, sufficient to exercise the schema-validated YAML write path but not the cloud-side `Open()` step.

**15 of 15 functional cases passed:**

| # | Case |
|---|---|
| 1 | `aws-login static` required only |
| 2 | `aws-login static --session-token` |
| 3 | `aws-login static --path other.creds` (writes at non-default path; preserves siblings) |
| 4 | `aws-login static --create` on already-existing env (no-op create) |
| 5 | `aws-login static --create` on a non-existent env (creates + writes in one command) |
| 6 | `aws-login static` against a non-existent env without `--create` (friendly error, non-zero exit) |
| 7 | `aws-login oidc` required only |
| 8 | `aws-login oidc` with `--duration --policy-arn ×2 --subject-attribute ×2` |
| 9 | `azure-login static` with required `<client-secret>` positional |
| 10 | `azure-login oidc` required only (new positional order) |
| 11 | `azure-login oidc --subject-attribute ×2` |
| 12 | `gcp-login static` required only |
| 13 | `gcp-login static --service-account --token-lifetime` |
| 14 | `gcp-login oidc` required flags only |
| 15 | `gcp-login oidc --region --token-lifetime --subject-attribute ×2` |

**Extra checks:**
- **Diagnostic rendering.** `aws-login oidc --duration not-a-duration` produced the rich hcl-rendered error with the source-line excerpt highlighting the offending YAML, the regex requirement from the backend, and a non-zero exit code. Confirms the diagnostic-handling change works end-to-end.
- **Direct API confirmation.** A YAML written by the CLI was fetched directly from `GET /api/esc/environments/<org>/<project>/<env>` and matched the CLI's `esc env get --definition` view byte-for-byte — the CLI is writing exactly what the backend stores.

**One bug surfaced and fixed during e2e (commit included in this PR):**
The shared `ensureProviderEnv` helper used by `--create` originally treated any non-nil error from the existence check as fatal, but the existence check returns an error for 404 (the very case `--create` is meant to handle). Fixed by short-circuiting `client.IsNotFound(err)` as "does not exist, proceed with create." After the fix, all 15 cases pass; commit on this branch.

## Test plan

- [x] Unit tests for each static + OIDC YAML node builder covering required-only, with-optional-fields, and with-array-fields cases.
- [x] Unit tests for shared `mergeProviderIntoEnv` helper: empty doc, replace-existing, preserve-siblings, plus `secretNode` tag enforcement.
- [x] Golden-file end-to-end tests for each provider × mode under `cmd/esc/cli/testdata/env-provider-*-login-{static,oidc}.yaml` (6 files).
- [x] Golden-file test for `--create` + missing-env behavior at `cmd/esc/cli/testdata/env-provider-missing-env.yaml`.
- [x] CLI output cross-checked against the documented provider schemas.
- [x] End-to-end run against a live backend (see results section above).
- [x] `make format && make lint && make test` all clean.


cmd/esc/cli/testdata/env-open-request-ok.yaml
```diff
@@ -1,6 +1,4 @@
-run: |
-  esc env open-request test-org/test-project/test-env
-  esc env open-request test-org/test-project/test-env --grant-expiration-seconds 3600s --access-duration-seconds 7200s  
+run: "esc env open-request test-org/test-project/test-env\nesc env open-request test-org/test-project/test-env --grant-expiration-seconds 3600s --access-duration-seconds 7200s  \n"
```
julienp: [nonblocking] some unnecessary reformatting here

cmd/esc/cli/env_provider.go
```diff
@@ -0,0 +1,25 @@
...
+)
+
+func newEnvProviderCmd(env *envCommand) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "provider",
+		Short: "Manage login providers within an environment",
+		Long: "Manage login providers within an environment\n" +
+			"\n" +
+			"Subcommands add cloud-provider login blocks (AWS, Azure, GCP) to an environment.\n" +
+			"Only static credentials are supported today; OIDC support will be added later.\n",
```
julienp: I think this is outdated

```suggestion
			"Subcommands add cloud-provider login blocks (AWS, Azure, GCP) to an environment.",
```

docs/superpowers/plans/2026-05-14-env-provider-login.md
```diff
@@ -0,0 +1,1132 @@
+# `esc env provider` Login Providers Implementation Plan
```
julienp: remove, I think, or does esc commit these things to the repo?

julienp (commented):
I think this looks ok in terms of flag naming (ref pulumi CLI), but I don't have the context to meaningfully review the functionality.

cmd/esc/cli/testdata/env-open-request-ok.yaml
borisschlosser: fixed

cmd/esc/cli/env_provider.go
borisschlosser: fixed

docs/superpowers/plans/2026-05-14-env-provider-login.md
borisschlosser: woops -> removed

cmd/esc/cli/env_provider.go
```diff
@@ -0,0 +1,22 @@
...
+package cli
+
+import (
+	"github.com/spf13/cobra"
+)
+
+func newEnvProviderCmd(env *envCommand) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "provider",
+		Short: "Manage login providers within an environment",
```
tehsis, line 12: Lets also point to the docs: https://www.pulumi.com/docs/esc/integrations/  / https://www.pulumi.com/docs/esc/integrations/dynamic-login-credentials/

cmd/esc/cli/env_provider_aws_login.go
```diff
@@ -0,0 +1,237 @@
...
+	"github.com/spf13/cobra"
+	"gopkg.in/yaml.v3"
+
+	"github.com/pulumi/pulumi/sdk/v3/go/common/resource"
+)
+
+func newEnvProviderAWSLoginCmd(env *envCommand) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "aws-login",
+		Short: "Add an AWS login provider to an environment",
```
tehsis, line 18: Add link to docs https://www.pulumi.com/docs/esc/integrations/dynamic-login-credentials/aws-login/ (same applies for other providers)

cmd/esc/cli/env_provider_aws_login.go
```diff
@@ -0,0 +1,237 @@
...
+		)
+	}
+	if len(subjectAttributes) > 0 {
+		oidcContent = append(oidcContent,
+			&yaml.Node{Kind: yaml.ScalarNode, Tag: "!!str", Value: "subjectAttributes"},
+			stringSequenceNode(subjectAttributes),
+		)
+	}
+
+	return &yaml.Node{
```
tehsis, line 237: Can we add a comment with the docs as well? (same for all providers)

tehsis (approved):

cmd/esc/cli/env_provider.go
borisschlosser: Done — added links to https://www.pulumi.com/docs/esc/integrations/ and https://www.pulumi.com/docs/esc/integrations/dynamic-login-credentials/ in the `provider` command's Long help.

cmd/esc/cli/env_provider_aws_login.go
borisschlosser: Done — added the per-provider docs link to the parent + `static` + `oidc` Long help text for `aws-login`, `azure-login`, and `gcp-login`.

PR #660, opened 2026-05-29
cli: add agent credential fallback

dschaller:
## Summary

For context this builds off https://github.com/pulumi/pulumi/pull/23225

Adds ESC CLI compatibility with Pulumi ephemeral agent credentials. ESC now honors `PULUMI_API`, falls back to Pulumi’s shared agent credential store when default credentials are unavailable in agent mode, and uses the Pulumi login manager so
agent accounts can be provisioned automatically.

This also upgrades Pulumi dependencies from `v3.209.0` to `v3.243.0` because the required agent credential APIs, including `GetAgentStoredCredentials` and `GetAccountWithAgentFallback`, are not available in the older version. That dependency bump requires updating ESC’s Go toolchain from 1.24 to 1.25.8.

Also propagates Cobra command contexts through ESC command handlers so embedded `pulumi env ...` commands preserve Pulumi’s agent credential usage tracking and can print the required claim/account instructions.

## Test changes

  Added coverage for agent-mode credential fallback, `PULUMI_API` backend selection, and `PULUMI_BACKEND_URL` precedence over `PULUMI_API`.

## Risk

Moderate. This changes credential/backend selection for agent environments and updates Pulumi dependencies. The intended blast radius is limited to detected agent runs without explicit PULUMI_CREDENTIALS_PATH or PULUMI_HOME; normal user credential resolution should continue using existing files.

Possible issues include selecting the wrong backend if env precedence is wrong, or unexpected behavior in self-hosted/agent combinations.

## Rollback

Revert the commits. This will restore prior ESC credential behavior and remove automatic agent credential fallback/context propagation.


cmd/esc/cli/workspace/credentials.go
```diff
@@ -80,18 +88,35 @@ func (w *Workspace) GetCurrentCloudURL(account *Account) string {
 	return "https://api.pulumi.com"
 }
 
+// TODO: Replace this with workspace.AgentCredentialsFallbackEnabled once ESC
+// depends on a Pulumi SDK release that includes it.
```
borisschlosser, line 92: Is that still a TODO with the pulumi dependency update?

borisschlosser (approved):

cmd/esc/cli/workspace/credentials.go
dschaller: Yeah once the pulumi sdk lands I'll update this

PR #665, opened 2026-06-11
Add support for environment overrides in ESC API

seanyeh:
related: https://github.com/pulumi/pulumi/pull/23562

### DIFF
--- a/cmd/esc/cli/client/client.go
+++ b/cmd/esc/cli/client/client.go
@@ -47,6 +47,10 @@ type CheckYAMLOption struct {
 	ShowSecrets bool
 }
 
+type OpenYAMLOption struct {
+	EnvironmentOverrides map[string]string
+}
+
 // Client provides a slim wrapper around the Pulumi HTTP/REST API.
 //
 // NOTE: this is not considered a public API, and we reserve the right to make breaking changes, including adding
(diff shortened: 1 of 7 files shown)


cmd/esc/cli/client/client.go
```diff
@@ -1057,13 +1062,24 @@ func (pc *client) OpenYAMLEnvironment(
...
 		Duration: duration.String(),
 	}
 
+	overrides := firstOrDefault(opts).EnvironmentOverrides
+	if len(overrides) > 0 {
+		encoded, err := json.Marshal(overrides)
+		if err != nil {
+			return "", nil, fmt.Errorf("marshaling environment overrides: %w", err)
+		}
+		queryObj.EnvironmentOverrides = string(encoded)
```
borisschlosser: Is there a specific reason why we are sending an encoded json as query parameter instead of a json body?

seanyeh: @borisschlosser for anonymous yaml open, we already send the environment definition as the body (as raw yaml), which makes it tricky to add anything else to the body. for the regular open, i kept the same for consistency

borisschlosser (changes requested):
Waiting for backend discussions to resolve: https://github.com/pulumi/pulumi-service/pull/44449

PR #669, opened 2026-07-06
Print retirement banner on every esc command

borisschlosser:
## What

With [pulumi/pulumi#23746](https://github.com/pulumi/pulumi/pull/23746) the standalone `esc` binary is being retired — users should use the Pulumi CLI's `pulumi env` subcommand instead. This adds a retirement notice printed on **every** invocation of the standalone `esc` binary, ahead of one final release.

```
┌─────────────────────────────────────────────────────────────────────────┐
│ The standalone ESC CLI has been retired and no longer receives updates. │
│ ESC is now part of the Pulumi CLI — use 'pulumi env' instead.           │
│ Install: https://www.pulumi.com/docs/install/                           │
└─────────────────────────────────────────────────────────────────────────┘
```

The wording scopes the retirement to the *binary* rather than the ESC product, and links straight to the install page — the actionable target, since commands map 1:1 (`esc env ls` → `pulumi env ls`).

## How

The banner is printed from `cmd/esc/main.go` (the `package main` entrypoint), to **stderr**, before `cli.New(nil).Execute()`.

Placing it in `main.go` rather than the shared `cli.New`/command tree means:

- It appears **only** in the standalone `esc` binary — never under `pulumi env`, however the Pulumi CLI now embeds the esc code.
- It prints for every command, including `--help`, `version`, and unknown/invalid commands.
- It goes to stderr, so it never corrupts machine-readable stdout (`esc env open`, `esc run`, `esc gen-docs`, ...).
- The 102 `cmd/esc/cli` stderr snapshot tests (which drive `New()` directly, not `main()`) are untouched.

The box auto-sizes to its longest line, so it stays aligned if the text changes.

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -8,7 +8,7 @@
 
 - Print a retirement notice on every `esc` command. The standalone ESC CLI has been retired;
   install the Pulumi CLI and use the `pulumi env` subcommand instead. See
-  https://www.pulumi.com/docs/iac/cli.
+  https://www.pulumi.com/docs/install.
   [#669](https://github.com/pulumi/esc/pull/669)
 
 ### Bug Fixes
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -6,6 +6,11 @@
 - Report the new revision number when an environment definition is updated (e.g. `esc env edit`,
   `esc env provider`, `esc env version rollback`).
 
+- Print a retirement notice on every `esc` command. The standalone ESC CLI has been retired;
+  install the Pulumi CLI and use the `pulumi env` subcommand instead. See
+  https://www.pulumi.com/docs/iac/cli.
+  [#PR](https://github.com/pulumi/esc/pulls/PR)
+
 ### Bug Fixes
 
 ### Breaking changes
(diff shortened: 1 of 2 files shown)


tehsis (approved):

djperlovsky:
Can we try the following? 

```
┌──────────────────────────────────────────────────────────────────────────┐
│ The standalone ESC CLI has been retired and no longer receives updates.  │
│ ESC is now part of the Pulumi CLI — use 'pulumi env' instead.            │
│ Install: https://www.pulumi.com/docs/install/                            │
└──────────────────────────────────────────────────────────────────────────┘
```

The current banner has three issues:

1. **"The ESC CLI has been retired"** reads like the ESC *product* is retired — only the standalone binary is. It should say **"standalone"**, matching the CHANGELOG entry.
2. **"Download the Pulumi CLI"** — "Install" is the standard verb in Pulumi docs, and many users already have `pulumi` installed.
3. **The link** (`/docs/iac/cli`) is the generic IaC CLI overview — it doesn't mention ESC, `pulumi env`, or migration. The install page is the actionable target.


"Standalone" scopes the retirement correctly, "now part of the Pulumi CLI" reassures users nothing is lost (commands map 1:1: `esc env ls` → `pulumi env ls`), and the link goes straight to the fix. The CHANGELOG_PENDING.md entry should link `/docs/install/` as well.

